Issue #237, opened 2017-04-20
Normative CSS2 reference changed to CSS2.1 in Second Edition


palemieux:
The reference to CSS2 in Second Edition changed from:

> Bert Bos et al., Cascading Style Sheets, Level 2, W3C Recommendation, 12 May 1998. (See http://www.w3.org/TR/1998/REC-CSS2-19980512/.)

to 

> Bert Bos et al., Cascading Style Sheets, Level 2 Revision 1, W3C Recommendation, 07 June 2011. (See http://www.w3.org/TR/2011/REC-CSS2-20110607/.)

Two issues:

- there were normative changes between the two CSS versions, e.g. "16.6.1 The 'white-space' processing model" introduced in CSS2.1
- Second Edition continues to reference XSL 1.1, which references CSS2

Proposed resolution:

- revert normative reference to CSS2

skynavga:
There is no bug here. TTML1 2nd Ed references CSS2.1. TTML1 also references XSL-FO 1.1, which references CSS2. So depending on the route to the specific reference, either CSS2.1 or CSS2 applies.

skynavga:
Committed in #238.

PR #238, opened 2017-04-25
Issue 0237 css2 reference

skynavga:
Downgrade CSS2 reference to informative since there was effectively no normative use. This also removes a potential reading ambiguity regarding which version of CSS is referenced, i.e., XSL-FO 1.1 references the original CSS2, while the [CSS2] reference in TTML refers to the newer CSS2.1.


palemieux:
I am not sure why CSS2.1 needs to be referenced, but perhaps it is already too late since Second Edition was published with that reference.

LGTM anyhow.

nigelmegitt (approved):
Looks good to me.

palemieux (approved):

skynavga:
Approved for merge in WG.

skynavga:
See #237.

PR #257, opened 2017-11-14
Tweak examples of Section O

### LINKED ISSUE #225: Paint-On example improvement.
The example in O.3 (Paint-On) has a couple problems in addition to #223 :

1. A region is defined, but not referenced. Add to div, or delete and make it clear it is using the root container.
~~2. There are no spaces between the words.  Add xml:space and appropriate spaces.~~ (per https://github.com/w3c/ttml1/issues/225#issuecomment-268376595)

### LINKED ISSUE #254: Examples in section O use an unusual tts:displayAlign setting
The examples in this section O are specifically named from the modes of CTA 608 closed captioning which implies certain behavior for the 3 modes. These 3 examples all take the initial value ("before") of tts:displayAlign, which causes initial text to appear in the top of the region. To align with 608 behavior it should be "after". Add tts:displayAlign="after" to the region definitions. (This difference is particularly noticeable for roll-up and paint-on.)

palemieux:
Close #225, close #254


spec/examples/ex2-pop-on.xml
```diff
@@ -2,8 +2,8 @@
   xmlns:ttp="http://www.w3.org/ns/ttml#parameter" xmlns:tts="http://www.w3.org/ns/ttml#styling">
   <head>
     <layout>
-      <region xml:id="r1" tts:color="white" tts:origin="10c 4c" tts:extent="40c 1c"/>
-      <region xml:id="r2" tts:color="yellow" tts:origin="10c 8c" tts:extent="40c 1c"/>
+      <region xml:id="r1" tts:color="white" tts:origin="10c 4c" tts:extent="40c 1c" tts:displayAlign="after"/>
+      <region xml:id="r2" tts:color="yellow" tts:origin="10c 8c" tts:extent="40c 1c" tts:displayAlign="after"/>
```
nigelmegitt: I don't think this is actively harmful but it is unnecessary - the region's extent is only `1c` high. Adding this detail could be confusing.

nigelmegitt (approved):
Just one comment - it's not bad enough to request a change per se, but @palemieux and @mikedo might want to have a look at the pop on example and consider if it is really helpful to put `tts:displayAlign` on a 1 line region.

palemieux:
> I don't think this is actively harmful but it is unnecessary - the region's extent is only 1c high. Adding this detail could be confusing.

Yes. I also wondered if a `1c` height is ever a good idea: it seems to be susceptible to font size changes, etc...

nigelmegitt:
> if a 1c height is ever a good idea

I'm fine with it in an example whose purpose is to show something else.

nigelmegitt (changes requested):
Looks good to me. Something weird is happening with the HTML generation, but the xml source looks good.

nigelmegitt (approved):

nigelmegitt:
Apologies for the confusion. I hit the Request changes button instead of approve back there, and then changed it to an approve.

skynavga (approved):
Doesn't need to rebuild schemas, but no harm perhaps.

nigelmegitt:
The built version of this PR can be viewed at https://rawgit.com/w3c/ttml1/issues-225-254-section-o-examples-build/index.html

PR #259, opened 2017-11-18
Clarify zIndex stacking context

### LINKED ISSUE #241: tts:zIndex does not specify the root container zIndex value when "auto" is not specified
Section 8.2.25 specifies that _the semantics of the value auto are those defined by [XSL 1.1], § 7.30.18, where the tt element is considered to establish the root stacking context_.

Does the `tt` element establish the stacking context even when `tts:zIndex` is not `auto`?

What is the computed `tts:zIndex` value for the root container? Does a region with `tts:zIndex < 0` appear underneath the root container?

palemieux:
Close #241


spec/ttml1.xml
```diff
@@ -6074,7 +6073,7 @@ the computed value, then the value least distant from 0, i.e., closest to the ba
 <note role="derivation">
 <p>The semantics of the style property
 represented by this attribute are based upon that defined by <bibref ref="xsl11"/>,
-&sect; 7.30.18.</p>
+&sect; 7.30.18, where the <el>tt</el> element is considered to establish the root stacking context.</p>
```
nigelmegitt: This doesn't map to TTML2 where it would conflict with w3c/ttml2#470 and it also moves an effectively normative statement into an informative one. I'd suggest moving the statement to a separate `p` just before or after the one about treatment of the `auto` value.

(it's a separate question if we want to back-port the CSS mappings into TTML1 Third Ed)

nigelmegitt (changes requested):

spec/ttml1.xml
```diff
@@ -5987,9 +5987,10 @@ as a style property only to those element types indicated in the following table
 overlap in space, the area(s) generated by lexically subsequent elements must be
 rendered over area(s) generated by lexically prior elements, where lexical
 order is defined as the postorder traversal of a <emph>Document Instance</emph>.</p>
+<p>The <el>tt</el> element is considered to establish the root of the stacking context, as defined by 
+<bibref ref="xsl11"/>, &sect; 7.30.18.</p>
 <p>The semantics of the value <code>auto</code> are those defined by
```
skynavga: With the proposed change, this paragraph can now be removed, since the definition of auto is already implied by the derivation note.

spec/ttml1.xml
```diff
@@ -6074,7 +6075,7 @@ the computed value, then the value least distant from 0, i.e., closest to the ba
 <note role="derivation">
 <p>The semantics of the style property
 represented by this attribute are based upon that defined by <bibref ref="xsl11"/>,
-&sect; 7.30.18.</p>
+&sect; 7.30.18</p>
```
skynavga: Need to restore the final PERIOD.

skynavga (changes requested):

spec/ttml1.xml
palemieux: @skynavga @nigelmegitt argues that the derivation note is informative and therefore cannot replace the normative clause _the semantics of the value auto are those defined by [XSL 1.1], § 7.30.18_

skynavga: I do not agree, because elsewhere, we do not normatively define in prose every keyword, but instead, rely upon the semantics of the derivation.

nigelmegitt: I'm arguing for minimal changes here. The reference to XSL was already present here, but it was not correct. Merely editing it so that the scope is correct is the change needed to address this issue. If you want to make another change to modify the way that the semantic derivation is stated then that's another separate issue.

If you want to raise that [other issue], then consider that this reference appears to be normative whereas the derivation statement is in a note. I would also request that any such change is made broadly in line with w3c/ttml2#470.

nigelmegitt: Having said that, I can see @skynavga 's point, and actually removing the statement about the semantics of `auto` altogether would not be harmful - I would not object to doing that, or to keeping it in.

nigelmegitt (approved):
Looks good to me as is. I would also not object to removing the statement about the semantics of `auto`.

spec/ttml1.xml
palemieux: @nigelmegitt How would the reader determine the semantics of `auto` if the clause _The semantics of the value auto are those..._ is removed?

nigelmegitt: @palemieux the same way the reader determines the semantics of `<integer>`!

palemieux: @nigelmegitt Alright.

palemieux: @skynavga Please approve the PR.

nigelmegitt (approved):

skynavga (approved):

nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml1/issue-241-clarify-zIndex-stacking-context-build/index.html

PR #260, opened 2017-11-19
Clarify that the computed value of tts:lineHeight='normal' is 'normal'

### LINKED ISSUE #220: Computed value of lineHeight and inheritance.
Section 8.4.4.2(6)(b) specifies that `lineHeight` is set to its initial value of `"normal"` on `<region>`.

Section 8.4.4.2(6)(c) specifies that `<body>` will inherit the _computed value_ of `lineHeight` from its parent `<region>`.

The computed value of `lineHeight` on `<region>` is therefore, say, 125% of the computed value of `fontSize` on `<region>`, so `125/15rh`.

So the computed values of `lineHeight` and `fontSize` on `<body>` are `125/15rh` and `400/15rh`, respectively.

These computed values are then propagated down by inheritance to the `<p>`, where "Hello" and "Bye" will overlap since the font size is significantly larger than the line height.

Is this the intended outcome and correct interpretation, or alternatively should initial values only be applied to elements to which the styles applies or does "normal" have special inheritance rules?

```
<?xml version="1.0" encoding="UTF-8"?>
<tt xmlns="http://www.w3.org/ns/ttml" xmlns:tts="http://www.w3.org/ns/ttml#styling">
    <head>
        <layout>
            <region xml:id="bottom" tts:origin="10% 10%" tts:extent="80% 80%" />
        </layout>
    </head>
    <body tts:fontSize="4c">
        <div>
            <p region="bottom" begin="00:00:10.000" end="00:00:20.000">Hello<br/>Bye</p>
        </div>
    </body>
</tt>
```

NOTE: `100rh` equals the computed height of root container.

palemieux:
Closes #220


spec/ttml1.xml
```diff
@@ -4594,26 +4594,26 @@ as a style property only to those element types indicated in the following table
 </tr>
 </tbody>
 </table>
-<p>If the value of this attribute is <code>normal</code>, then the
-computed value of the style property must be considered to be no less than the
+<p>If the computed value of this attribute is <code>normal</code>, then the
+used value of the style property must be considered to be no less than the
```
nigelmegitt: Interested to know if anyone else thinks we need to state that the terms  "specified value", "computed value", "used value" and "actual value" are as defined in [CSS](https://www.w3.org/TR/CSS2/cascade.html#value-stages)?

spec/ttml1.xml
```diff
@@ -4594,26 +4594,26 @@ as a style property only to those element types indicated in the following table
...
+<p>If the computed value of this attribute is <code>normal</code>, it is recommended that presentation processors
+use a line height that is 125% of the size of the largest applicable font size.
+The choice of 125% allows use of a font size of <code>0.8c</code> in order
 to obtain a line height of <code>1c</code>.</p>
 </note>
 <p>If specified as a <loc
 href="#style-value-length">&lt;length&gt;</loc>, then the length must
 be non-negative.</p>
-<p>If a computed value of the property associated with this attribute is not supported,
+<p>If the used value of the property associated with this attribute is not supported,
```
nigelmegitt: Could reword in terms of the actual value, since it looks odd to say "if the used value is not supported, use a different value" (both are expressed in terms of "use", whereas in fact it's the mapping from used value to actual value that is being affected, in CSS terms)

spec/ttml1.xml
```diff
@@ -4594,26 +4594,26 @@ as a style property only to those element types indicated in the following table
 </tr>
 </tbody>
 </table>
-<p>If the value of this attribute is <code>normal</code>, then the
```
nigelmegitt: Looking at the [CSS definition](https://www.w3.org/TR/REC-CSS2/visudet.html#line-height), it is clear that `normal` is treated as ` <number>`, and that when `<number>` is specified, the _number_ not the computed value is inherited. This is not clear either in XSL or in the table above. I think we need a specific line of text to clarify it, e.g.:

> When the specified value is "normal" it is the specified value that is inherited, not the computed value.

nigelmegitt (changes requested):
Doesn't quite address the issue for me - my suggestion is to state explicitly that "normal" is the inherited value, not the computed value.

palemieux:
@nigelmegitt I have updated the PR to remove the need to define "used value" and more closely align the prose with XSL 1.1 and CSS 2.1, which note that _In most cases, elements inherit computed values. However, there are some properties whose specified value may be inherited (e.g., the number value for the 'line-height' property). In the cases where child elements do not inherit the computed value, this is described in the property definition._

spec/ttml1.xml
```diff
@@ -4598,22 +4598,22 @@ as a style property only to those element types indicated in the following table
...
-is specified. The choice of 125% allows use of a font size of <code>0.8c</code> in order
+<p>If the value of this attribute is <code>normal</code>, it is recommended that presentation processors
+use a line height that is 125% of the size of the largest applicable font size.
+The choice of 125% allows use of a font size of <code>0.8c</code> in order
 to obtain a line height of <code>1c</code>.</p>
 </note>
 <p>If specified as a <loc
 href="#style-value-length">&lt;length&gt;</loc>, then the length must
 be non-negative.</p>
-<p>If a computed value of the property associated with this attribute is not supported,
```
skynavga: Change is gratuitous and unnecessary. Please restore 'a' to be consistent with language elsewhere.

spec/ttml1.xml
```diff
@@ -4598,22 +4598,22 @@ as a style property only to those element types indicated in the following table
...
+<p>If the computed value of the property associated with this attribute is not supported,
 then a presentation processor must use the closest supported value.</p>
 <note role="elaboration">
 <p>In this context, the phrase <emph>closest supported value</emph> means the value for which the Euclidean distance between
-the computed line height and the supported line height is minimized. If there are multiple closest supported values equally distant from
```
skynavga: Other than source formatting (removal of a newline), there seems to be no change here. Please restore the newline in which case I believe this change is a no-op (unless there is something subtle I don't see here).

spec/ttml1.xml
```diff
@@ -4598,22 +4598,22 @@ as a style property only to those element types indicated in the following table
 computed value of the style property must be considered to be no less than the
 size of the largest font size that applies to the element and its
 descendant elements in the intermediate synchronic document as determined by
-<specref ref="semantics-region-layout-step-1"/>.</p>
+<specref ref="semantics-region-layout-step-1"/>. However, the value <code>normal</code> is not relative,
+and is therefore inherited as-is, as specified at <specref ref="semantics-style-resolution-processing-css"/>.</p>
 <note role="elaboration">
-<p>It is recommended that presentation processors use a computed line height that is
```
skynavga: Please restore the original text, as this editorial change is gratuitous and unnecessary.

spec/ttml1.xml
```diff
@@ -4598,22 +4598,22 @@ as a style property only to those element types indicated in the following table
 computed value of the style property must be considered to be no less than the
 size of the largest font size that applies to the element and its
 descendant elements in the intermediate synchronic document as determined by
-<specref ref="semantics-region-layout-step-1"/>.</p>
+<specref ref="semantics-region-layout-step-1"/>. However, the value <code>normal</code> is not relative,
```
skynavga: This additional sentence is unclear and confusing in its introduction of the term 'relative' out of the blue as it were. We need to reconsider what we are trying to do/say here then find a better formulation for doing so.

spec/ttml1.xml
```diff
@@ -7107,6 +7107,9 @@ context of a layout or <loc href="#element-vocab-type-content">Content</loc> ele
 context, and not in the non-presentation, declaration context
 of a referentiable <el>style</el> element.</p>
 </note>
+<note role="clarification">
```
skynavga, line 7139: This addition is also problematic and not adequately motivated or contextualized. Need to reconsider as suggested above.

skynavga (changes requested):

spec/ttml1.xml
palemieux: @skynavga The term "relative" is key since 8.4.4.3(4) specifies that only relative value types, e.g. "%" or "em", have their specified value replaced by their computed values in the _ Computed Style Set_, which is inherited in 8.4.4.2(c). In this case, we do not want "normal" to be replaced by its computed value, so it is sufficient to clarify that it is not a relative value.

spec/ttml1.xml
palemieux: The current text does not explicitly associate the note with the condition that "the value of this attribute is <code>normal</code>"

spec/ttml1.xml
palemieux: It is unavoidable that whitespaces will be inserted in the editing process. One way to avoid this is to agree on a formatting tool that breaks lines consistently.

spec/ttml1.xml
palemieux: The idea is to remind the reader that the value "normal" is inherited, instead of its computed value, since it is not a relative value type.

spec/ttml1.xml
skynavga: No change should be made by the editing tool if there is no change in the source content. What is the change here that induced the diff delta? Also, no gratuitous (non-explicit) change in source whitespace should occur due to whatever editing tool is being used.

nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml1/issue-220-lineheight-normal-clarification-build/index.html

spec/ttml1.xml
cconcolato: I don't see any text specific to line height in the CSS inheritance rule: https://www.w3.org/TR/css3-cascade/#inheriting

palemieux: @cconcolato See last paragraph of Section 5.1.2 at https://www.w3.org/TR/xsl11/

spec/ttml1.xml
skynavga: In that case, you could have avoided changing the phrase order of the sentence and just substituted "when the value of this attribute is ``<code>normal</code>``" for the original phrase "when the value ``<code>normal</code>`` is specified".

In other words, please attempt to minimize changes by avoiding a wholesale rewrite of a sentence when only a small delta will suffice; that is, unless the original sentence is completely broken.

spec/ttml1.xml
```diff
@@ -7107,6 +7107,9 @@ context of a layout or <loc href="#element-vocab-type-content">Content</loc> ele
...
+<p>Step (4) is not applied for the style property <code>["http://www.w3.org/ns/ttml#styling", "lineHeight"]</code> when its value expression is <code>normal</code>, which is not considered relative.</p>
```
skynavga: This note has to be deleted. We can't (and shouldn't) list (or take note of) exceptional processing regarding what is or what is not relative in this context. If it is felt that a note is absolutely necessary, then it should say nothing more than "See individual style properties for information regarding what values are considered to be relative for the purpose of interpreting step (4) above."

spec/ttml1.xml
skynavga: I think we have more of a problem than I initially suspected since the text here (and TTML2) makes normative statements about resolving ``normal`` to an absolute value in the defining context (rather than the usage context). We may need to introduce the category *used values* after all.

skynavga:
@nigelmegitt I think we cannot introduce special inheritance semantics here, as that would be very close (if not precisely) a normative change. Rather, we need to craft the minimal changes needed to employ the default inheritance logic, which is the inheritance of computed values.

spec/ttml1.xml
palemieux: @skynavga Happy to use _See individual style properties for information regarding what values are considered to be relative for the purpose of interpreting step (4) above._

Would that address all your comments?

skynavga: Just for this note, but not the other items. As I am preparing a PR against this PR, I will include this change.

nigelmegitt:
@skynavga I would argue that this change is a reformulation of this existing slightly odd statement in the spec:

> If the value of this attribute is normal, then the computed value of the style property must be considered to be no less than the size of the largest font size that applies to the element and its descendant elements in the intermediate synchronic document as determined by 9.3.2 Intermediate Synchronic Document Construction.

which better aligns it with the existing semantics of XSL 1.1 (and CSS 2) which arguably already apply, makes it easier to understand the origin and reasoning, and makes it easier to see how to implement it. In other words, I'd make the case that it is not a normative change.

Incidentally, you seem to imply that we are not free to make normative changes - why?

palemieux:
@nigelmegitt Are you happy with the PR as it stands? Please point out your concerns.

skynavga:
@nigelmegitt short of introducing the distinction of "used values", as in CSS2.1 (2011), I am now beginning to favor your suggestion that we call out exceptional inheritance behavior for ``normal``, which is to say (the equivalent of):

>The value ``normal`` is considered to be non-relative for the purpose of inheritance, but relative at the context of use; that is, the computed value used during inheritance is ``normal`` but is absolutized at the most nested elements to which this property applies.

I should have use the term *substantive* change. I was referring to either a change in syntax or an explicit change in semantics (from what was originally intended). I do not consider a change to normative prose that corrects or clarifies an ambiguity to constitute a *substantive* change.

nigelmegitt:
@palemieux thanks, I think you've addressed my comments, I'd like to wait for any further edits needed to resolve @skynavga 's comments before doing a final re-review. It seems that he is coming round to my point of view, but that he might ask for a rewording for clarity, precision or whatever.

andreastai:
It should be clarified that we try to solve issues in a PR in that PR and not in an alternative PR (as done with 1f438c5). Forking the discussion does not help at all and makes it very difficult to follow.

In general I agree with this PR. If there are concerns how the the term "relative" is used, this term may just be removed.

A code example may help to understand the inheritance mechanism.

nigelmegitt:
@tairt Sorry, I don't exactly agree. If there's a significant modification to a pull request then opening another pull request against it is reasonable. That way there's only one "main" pull request that can be merged. Alternatively if there's a completely independent different approach to resolving an issue, so that it is effectively a "competitor" to an existing pull request, then it should be done as a separate PR in its own right.

I agree that it's hard to follow though, and actually for simple changes it is probably easier just to add the proposed changed text into a comment, discuss it, and when there's a consensus, the originator of the pull request can make the changes directly into the pull request branch.

skynavga:
@tairt to add to @nigelmegitt comments, my doing a PR against the original PR was at the (private) request of @palemieux in order to avoid my changing the original PR directly (which I could have done by pushing a commit to the original PR); in other words, we did it this way to avoid it being an "alternative PR"

palemieux:
I have revised the PR with the following in mind:

- minimize changes to TTML1 [ed.: I suggest avoiding extensive additional semantics as proposed in https://github.com/w3c/ttml1/pull/277]

- match the approach taken with `tts:textDecoration` for specifying special inheritance semantics

- match the XSL approach, which uses the following prose in the definition of `line-height`: _The computed value of the property is this number multiplied by the element's font size. Negative values are illegal. However, the number, not the computed value, is inherited._

nigelmegitt (approved):
Looks good to me!

skynavga (approved):

andreastai (approved):

PR #262, opened 2017-11-22
Travis CI integration

palemieux:
* a complete specification is generated by 'ant build' under the build/ directory
* Travis stores the complete specification on a sister branch called ${current-branch-name}-build
* the `build` target expands keywords
* build artifacts are git-ignored


spec/build.xml
```diff
@@ -230,12 +230,30 @@
...
+        <delete includeemptydirs="true" quiet="true">
+          <fileset dir="${build.target}"/>
+        </delete>
+        <mkdir dir="${build.target}"/>
+        <mkdir dir="${build.target}/images"/>
+        <mkdir dir="${build.target}/rnc"/>
+        <mkdir dir="${build.target}/xsd"/>
+        <copy file="${html.dir}/${html.name}" tofile="${build.target}/index.html"/>
+        <copy todir="${build.target}/images">
+          <fileset dir="${basedir}/images" includes="*.png"/>
```
nigelmegitt: At least in TTML2 there are SVG images too

nigelmegitt (approved):
This is much better than what we had before, so I'm keen to merge it soon even if there are further tweaks that could be made.

I notice that it is possible to have build branch name clashes, since the same branch name can be reused (I know!), but this is unlikely to be a significant problem.

nigelmegitt (commented):
It's really hard to review this commit on the GitHub web page because it treats the indentation changes as huge blocks of removed and added lines. Is it just changing the indentation, or were there other changes too?

palemieux:
> Is it just changing the indentation, or were there other changes too?

There were other changes, including:

- relocating targets within the file
- removing targets that committed the built specification

`git diff -w` can be used to view the changes ignoring whitespace.

nigelmegitt:
Ah, thanks, that was helpful! All looks fine to me. What action is needed about the lines that look like:

```xml
<!-- DO NOT USE AS-IS -->
```
?

palemieux:
These targets need to be at the very least modified — since they have hardcoded pub names. I plan on getting to them as we get closer to publication. In the meantime they should not be used.

nigelmegitt:
OK, no problem, the build script doesn't use them anyway.

palemieux:
The build target does not use the sg and rg targets either. I thought I would keep mods to a minimum for now. The script will benefit from additional maintenance over time.

PR #264, opened 2017-11-23
Clarify semantics in absence of begin attribute

### LINKED ISSUE #224: No begin times in Pop-On example in Appendix O.

palemieux:
Closes #224


spec/ttml1.xml
```diff
@@ -7781,6 +7781,9 @@ interval; i.e., the interval is left-wise closed.</p>
 <p>The semantics of the <att>begin</att> attribute are those defined
 by <bibref ref="smil21"/>, &sect; 10.4.1,
 while taking into account any overriding semantics defined by this specification.</p>
+<note role="clarification">
```
skynavga, line 7790: Replace with following "See ``<bibref ref="smil21"/>``, &sect; 10.4.1, ``<emph>``Begin value semantics``</emph>``, for information about default timing."

skynavga (changes requested):

spec/ttml1.xml
nigelmegitt: I'm happy with either the pull request version or @skynavga 's proposed alternative.

Also as per https://github.com/w3c/ttml1/issues/224#issuecomment-268787562 it would be useful to state that TTML does not define any epoch or epoch offset, and that if those are relevant they need to part of the document processing context.

palemieux: > Also as per #224 (comment) it would be useful to state that TTML does not define any epoch or epoch offset, and that if those are relevant they need to part of the document processing context.

Please file a separate issue and propose specific text.

palemieux: @skynavga The original comment was behavior when `begin` is not specified, so I think the note needs to mention that. What about:

_See `<bibref ref="smil21"/>`, § 10.4.1, `<emph>Begin value semantics</emph>` when no `begin` attribute is specified`_

nigelmegitt: > Please file a separate issue and propose specific text.

#266

nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml1/issue-224-default-begin-build/index.html

palemieux:
@skynavga Please review and approve (as the case may be).

skynavga (approved):

nigelmegitt (approved):
Looks good to me.

PR #270, opened 2017-11-30
Clarified application of tts:visibility to child elements

### LINKED ISSUE #207: Clarify that tts:visibility hidden on ancestor doesn't prevent visible on descendant.
The current text on tts:visibility may be read such that if tts:visibility is hidden on an element, then all descendants are hidden, even if they specify tts:visibility as visible. This reading is not intended, so a note or additional clarifying text should be added to indicate that descendants may be visible.

palemieux:
Closes #207

Prose copied from [TTML2](https://www.w3.org/TR/ttml2/#style-attribute-visibility)


nigelmegitt (approved):
Looks good to me.

spec/ttml1.xml
```diff
@@ -5672,6 +5672,10 @@ each character child of a <el>p</el> element is considered to be enclosed in an
 span.</p>
 <p>The <att>tts:visibility</att> style has no affect on content layout or
 composition, but merely determines whether composed content is visible or not.</p>
+<p>If the computed value of the property associated with this attribute is visible, then areas generated
```
skynavga: The 1st and 5th instances of 'visible' should be wrapped in ``<code>...</code>``; likewise for the 1st instance of 'hidden'. This is to match the formatting in TTML2.

skynavga (changes requested):

palemieux:
@skynavga Please review

skynavga (approved):

PR #276, opened 2017-12-05
<set> elements no longer pruned by ISD

### LINKED ISSUE #216: Process [construct intermediate document] prunes "set" elements.
See https://github.com/w3c/ttml2/issues/210

palemieux:
Closes #216


skynavga (approved):

spec/ttml1.xml
```diff
@@ -7405,7 +7405,8 @@ headed by the <el>body</el> element;</p>
 </item>
 <item>
 <p>evaluating this sub-tree in a postorder traversal, prune elements if they
-are not a <loc href="#element-vocab-type-content">Content</loc> element, if they are temporally inactive, if they are empty,
+are not a <loc href="#element-vocab-type-content">Content</loc> or <loc href="#element-vocab-type-animation">Animation</loc> element, 
+if they are temporally inactive, if they are empty,
```
nigelmegitt: What does "if they are empty" mean in this context? Now that we are trying to avoid pruning `set` elements, I think we need to change this "if they are empty" clause as well. Looking at https://www.w3.org/TR/ttml1/#animation-vocabulary-set the only contents permitted on a `<set>` are metadata; consequently it is likely that well formed `<set>` elements that should do something would be considered empty and therefore would be pruned, in this case unintentionally.

One solution to this would be to change the phrase to "if they are an empty Content element". Another would be to adopt the TTML2 approach of defining [presentation related element](https://www.w3.org/TR/ttml2/#terms-presentation-related-element) and referencing that term. The term would have to be defined a little differently because some classes of element referenced by the term definition in TTML2 are not present in TTML1.

nigelmegitt (commented):
I've asked a question - it seems that a further change is probably needed to resolve the issue.

nigelmegitt (approved):
Thanks, looks good to me.

PR #277, opened 2017-12-07
Update for PR #260 to resolve #220.

### LINKED ISSUE #220: Computed value of lineHeight and inheritance.

skynavga:


spec/ttml1.xml
```diff
@@ -4598,12 +4598,19 @@ as a style property only to those element types indicated in the following table
 computed value of the style property must be considered to be no less than the
 size of the largest font size that applies to the element and its
 descendant elements in the intermediate synchronic document as determined by
-<specref ref="semantics-region-layout-step-1"/>. However, the value <code>normal</code> is not relative,
-and is therefore inherited as-is, as specified at <specref ref="semantics-style-resolution-processing-css"/>.</p>
+<specref ref="semantics-region-layout-step-1"/>. Furthermore,
+when performing <specref ref="semantics-style-resolution-processing-css"/>, step (4), [relative value resolution],
+the value <code>normal</code> is considered to be <emph>relative</emph> only on the most nested elements to which this property applies, namely,
+on <el>p</el> elements; otherwise, it is not considered to relative.</p>
 <note role="elaboration">
```
palemieux, line 4605: Why would the value `normal` ever be considered relative?

skynavga: When it is computed to be 125% of the font size of the paragraph; ergo, it is relative to the computed value of the font size of the paragraph.

palemieux: In TTML1, it is not 125% of the font size of the paragraph, it _must be considered to be no less than the size of the largest font size that applies to the element and its descendant elements in the intermediate synchronic document as determined by 9.3.2 Intermediate Synchronic Document Construction._

skynavga: more precisely, it is ``min(*what you say*,*whatever the implementation wants but is recommended to be 125% font size*)``, but, in any case, that still makes it relative on p, which is what you asked for

nigelmegitt: I disagree @skynavga , I think the specified value `"normal"` is _never_ relative, but does generate a computed line height that is indeed related to the font size.

skynavga: @nigelmegitt then your understanding is at variance with both (1) XSL-FO, which, e.g., considers ``auto`` (on any property) to be relative, and (2) the computed style set processing algorithm, which requires a value to be treated as relative in order to map its category to ``computed``

from XSL-FO 1.1 &#x00A7;5.1.2, we have (emphasis added)

>Specified values may be absolute (i.e., they are not specified relative to another value, as in "red" or "2mm") or relative (i.e., they are specified relative to another value, as in **"auto"**, "2em", and "12%"), or they may be expressions.

nigelmegitt: @skynavga would you not allow that `"normal"` is an expression then, and neither absolute nor relative?

skynavga: @nigelmegitt in this context, *expression* is referring to the XSL-FO expression language defined in https://www.w3.org/TR/xsl11/#d0e5032; so, while ``normal`` could potentially be interpreted as an EnumerationToken (in the XSL-FO expression language), that interpretation doesn't really apply in TTML;

the bottom line is that in TTML a style value can only be designated as a computed value in a computed style set when it is a relative value; otherwise, it is a specified value in a computed style set;

furthermore, the language I cited today from TTML1 &#x00A7;8.2.12, namely

>If the value of this attribute is normal, then the computed value of the style property must be ...

makes sense only when it is assigned the category ``computed``, and that can only happen if it is relative (per &#x00A7;8.4.4.3, step 4);

nigelmegitt: @skynavga okay, thanks for the explanation. I may be repeating part of the conversation on Thursday, but it seems like we need to be able to inherit `"normal"` as a specified value on elements where it does not _apply_ and only compute the value on elements where it does apply.

skynavga: yes, that is exactly what I've been trying to say

spec/ttml1.xml
```diff
@@ -4598,12 +4598,19 @@ as a style property only to those element types indicated in the following table
 computed value of the style property must be considered to be no less than the
 size of the largest font size that applies to the element and its
 descendant elements in the intermediate synchronic document as determined by
-<specref ref="semantics-region-layout-step-1"/>. However, the value <code>normal</code> is not relative,
-and is therefore inherited as-is, as specified at <specref ref="semantics-style-resolution-processing-css"/>.</p>
+<specref ref="semantics-region-layout-step-1"/>. Furthermore,
+when performing <specref ref="semantics-style-resolution-processing-css"/>, step (4), [relative value resolution],
+the value <code>normal</code> is considered to be <emph>relative</emph> only on the most nested elements to which this property applies, namely,
```
nigelmegitt, line 4603: I don't understand the function of the words "most nested" here. Can we drop them?

spec/ttml1.xml
```diff
@@ -4598,12 +4598,19 @@ as a style property only to those element types indicated in the following table
...
+on <el>p</el> elements; otherwise, it is not considered to relative.</p>
```
nigelmegitt, line 4604: typo: "to relative" -> "to be relative"

spec/ttml1.xml
nigelmegitt: OK, this works for me, modulo the two comments I just added about specific bits of the wording.

spec/ttml1.xml
```diff
@@ -4598,12 +4598,19 @@ as a style property only to those element types indicated in the following table
...
 <note role="elaboration">
-<p>If the value of this attribute is <code>normal</code>, it is recommended that presentation processors
-use a line height that is 125% of the size of the largest applicable font size.
-The choice of 125% allows use of a font size of <code>0.8c</code> in order
+<p>Since the value <code>normal</code> is considered to be non-relative for all ancestors of <el>p</el>,
+this value is inherited as is without further resolution; however, when the computed style set is obtained on
+a <el>p</el> element, the value <code>normal</code> is treated as a relative value, and, therefore, is resolved
+to an absolute form during [relative value resolution].</p></note>
+<note role="elaboration">
+<p>It is recommended that presentation processors use a computed line height that is
```
nigelmegitt, line 4611: I don't understand why this sentence has been reordered, and it seemed better before. Beginning with the condition helps define the scope better. "If this then that" is a good pattern.

nigelmegitt (commented):
The direction of this makes sense to me.

skynavga:
#220 has been resolved to my satisfaction

PR #279, opened 2017-12-12
Specify semantics of fontSize in 'em' on region

### LINKED ISSUE #206: Use of 'em' units in tts:fontSize on region element is not well defined.
§8.2.9 needs to place a constraint on the use of 'em' units when expressing tts:fontSize on a region element; alternatively, it could specify that 1em is equal to the initial value of tts:fontSize, namely, 1c, for the purpose of resolving an 'em' unit based font size

palemieux:
Close #206


spec/ttml1.xml
```diff
@@ -4303,6 +4303,7 @@ glyphs do not have an EM square that is distinct from their font's EM square.</p
 <p>If horizontal and vertical sizes are expressed independently, then the
 units of the <loc href="#style-value-length">&lt;length&gt;</loc> values must be the same.</p>
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
+<p><loc href="#style-value-length">&lt;length&gt;</loc> values specified on a <loc href="#content-vocabulary-span"><el>region</el></loc> element using <code>em</code> units are considered relative to the <emph>Computed Cell Size</emph>.</p>
```
nigelmegitt: Are they independently related to the cell height and width, or always to the cell height? For example, is `"1em 1em"` supposed to result in the same behaviour as `"100% 100%"` or as `"100%"`?

nigelmegitt (commented):
Question asked

spec/ttml1.xml
palemieux: I think this is covered by the note:

_The expression 1c means one cell, where 'c' expresses the cell length unit as defined by 8.3.9 <length>. When a single <length> is expressed using cell units, then it refers to the height of the Computed Cell Size. When two <length> values are expressed using cell units, then the first refers to the width of the Computed Cell Size, and the second refers to the height of the Computed Cell Size._

nigelmegitt: @palemieux that note explicitly refers to the use of cell units not `em` units so I don't think it does cover it.

nigelmegitt: By the way, there's another relevant paragraph at https://www.w3.org/TR/ttml1/#style-value-length :

> When specified relative to a font whose size is expressed as a single length measure or as two length measures of equal length, the unit of measure em is considered to be identical to that defined by [XSL 1.1], § 5.9.13; however, when specified relative to a font whose size is expressed as two length measures of non-equal lengths, then one em is equal to the inline progression dimension of the anamorphically scaled font when used to specify lengths in the inline progression direction and equal to the block progression dimension of the scaled font when used to specify lengths in the block progression direction.

However this doesn't really help when the words "relative to a font" don't apply, as here, where the `em` measure is relative to the computed cell size.

palemieux:
@nigelmegitt See text at https://github.com/w3c/ttml1/pull/279/commits/51aa74f4ea3e2cfdf4a3adbf4f8843c1fcb8c72c

spec/ttml1.xml
```diff
@@ -4303,7 +4303,12 @@ glyphs do not have an EM square that is distinct from their font's EM square.</p
 <p>If horizontal and vertical sizes are expressed independently, then the
 units of the <loc href="#style-value-length">&lt;length&gt;</loc> values must be the same.</p>
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
-<p><loc href="#style-value-length">&lt;length&gt;</loc> values specified on a <loc href="#content-vocabulary-span"><el>region</el></loc> element using <code>em</code> units are considered relative to the <emph>Computed Cell Size</emph>.</p>
+<p><loc href="#style-value-length">&lt;length&gt;</loc> values specified on a <loc href="#content-vocabulary-span">
+<el>region</el></loc> element using <code>em</code>
```
nigelmegitt: the link points to span but the text says region - copy/paste error?

nigelmegitt (changes requested):
Thanks, the update addresses the comment, just noticed a typo to be fixed before approval.

palemieux:
@nigelmegitt Turns out the paste was correct, but the spec was incorrect where the copy was made :) Fixed now.

nigelmegitt (approved):
Thanks. Funny way to spot spec bugs, but whatever works!

spec/ttml1.xml
```diff
@@ -4303,6 +4303,12 @@ glyphs do not have an EM square that is distinct from their font's EM square.</p
 <p>If horizontal and vertical sizes are expressed independently, then the
 units of the <loc href="#style-value-length">&lt;length&gt;</loc> values must be the same.</p>
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
+<p><loc href="#style-value-length">&lt;length&gt;</loc> values specified on a <loc href="#layout-vocabulary-region">
+<el>region</el></loc> element using <code>em</code> 
+units are considered relative to the <emph>Computed Cell Size</emph>. When a single value is specified, then it is
```
skynavga: Is the second sentence of this paragraph "When a single ..." dependent on the context in the first sentence, i.e., does the second sentence only apply for length's using ``em``? Or is a general statement (all lengths)? If the former, then change the second sentence to start with "In that case, when a single ...".

spec/ttml1.xml
```diff
@@ -4303,6 +4303,12 @@ glyphs do not have an EM square that is distinct from their font's EM square.</p
 <p>If horizontal and vertical sizes are expressed independently, then the
 units of the <loc href="#style-value-length">&lt;length&gt;</loc> values must be the same.</p>
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
+<p><loc href="#style-value-length">&lt;length&gt;</loc> values specified on a <loc href="#layout-vocabulary-region">
```
skynavga: Need to tweak to prevent use of ``<length>`` at start of sentence, e.g., change to "When ``<length>`` values ...".

skynavga (changes requested):
See above comments around lines 4306-4308.

palemieux:
@skynavga Edits made

spec/ttml1.xml
```diff
@@ -4303,6 +4303,14 @@ glyphs do not have an EM square that is distinct from their font's EM square.</p
 <p>If horizontal and vertical sizes are expressed independently, then the
 units of the <loc href="#style-value-length">&lt;length&gt;</loc> values must be the same.</p>
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
+<p>When <loc href="#style-value-length">&lt;length&gt;</loc> values is specified on a <loc href="#layout-vocabulary-region">
```
skynavga: Needs a final tweak, either "When a ``<length>`` value is ..." or "When ``<length>`` values are ...".

skynavga (approved):
Thanks.

nigelmegitt (approved):

palemieux:
Propose to replace by #301, which is based on TTML2 prose

palemieux:
Mooted by merging of #301

PR #281, opened 2017-12-12
Clarify percentage units in padding specification

### LINKED ISSUE #205: Clarify meaning of percentage with writing mode relative edge terms in tts:padding.
Initially reported by Andreas Tai.

The question here is, for tts:padding="1% 2% 3% 4%" are the values related to absolute height and width or writing mode relative dimensions?

palemieux:
Close #205


spec/ttml1.xml
```diff
@@ -5059,6 +5059,9 @@ before edge, the second applies to the start and end edges, and the third applie
 to the after edge.
 If four <loc href="#style-value-length">&lt;length&gt;</loc> specifications are provided, then they apply to before, end,
 after, and start edges, respectively.</p>
+<note role="clarification">
+<p>Pecentage values are relative to the dimension of the region to which they apply. For example, if the before and after edges correspond to the top and bottom edges of the region then percentage values that apply to either of the two edges are relative to the height of the region.</p>
```
nigelmegitt: Looks good, even if I did have to read it three times. Would it be helpful to mention writing mode dependency explicitly to put the reader's mind in the right place for understanding that before and after are not always top and bottom?

nigelmegitt (approved):
Approving, though it could possibly be improved further - see comment.

skynavga (approved):

xchange11 (commented):
(mistakenly posted)

andreastai (commented):
I agree with @nigelmegitt that it would be good to give some more hints about the relevant aspects. It would be helpful to give examples for two different wirting directions e.g. in the text below:

"For example for a Latin script text with left-to-right-top-to-bottom writing mode the before and after edges correspond to the top and bottom edges of the region. In this case the percentage values that apply to before and after edges are relative to the height of the region. On the other hand for an Asian Script Text with a top-to-bottom-right-to-left writing mode the before and after edges correspond to the right and left edges of the region. In this case the percentage values that apply to before and after edges are relative to the width of the region."

spec/ttml1.xml
```diff
@@ -5071,7 +5071,7 @@ to the after edge.
 If four <loc href="#style-value-length">&lt;length&gt;</loc> specifications are provided, then they apply to before, end,
 after, and start edges, respectively.</p>
 <note role="clarification">
-<p>Pecentage values are relative to the dimension of the region to which they apply. For example, if the before and after edges correspond to the top and bottom edges of the region then percentage values that apply to either of the two edges are relative to the height of the region.</p>
+<p>Pecentage values are relative to the dimension of the region to which they apply. For example, if the before and after edges correspond to the top and bottom edges of the region, as is the case for Latin text where <att>tts:writingMode</att> is equal to <code>"lrtb"</code>, then percentage values that apply to either of the two edges are relative to the height of the region. Conversly, if the before and after edges correspond to the right and left edges of the region, as is the case for Japanese text where <att>tts:writingMode</att> is equal to <code>"tbrl"</code>, then percentage values that apply to either of the two edges are relative to the width of the region.</p>
```
skynavga: I do not agree with this change, as it is now attempting to explain (and paraphrase) the relevant XSL-FO semantics of what is meant by writing mode.

skynavga (changes requested):

spec/ttml1.xml
palemieux: @skynavga The proposed modification merely adds a second example. Were you not happy with the original text, which had a single example?

skynavga: I already accepted the original text (of line 5074 above shown in red). I do not accept the newest change that further elaborates the example.

nigelmegitt: @skynavga is there an error in the newer change with the further elaborated example? It's certainly easier to grasp, and since it is in a clarifying note, that seems like a good thing.

nigelmegitt (approved):
Additional examples look good to me, and make it easier to understand the point being made.

spec/ttml1.xml
skynavga: It now leaves the impression that it applies only to latin and japanese text. There is no need to explain writing mode in this context, and it produces clutter. Too much text is a bad thing and this example has now crossed the line.

nigelmegitt: OK in general over-explaining is unhelpful. I don't agree that it leaves the impression that it only applies to latin and japanese text though. Could you please propose something that elaborates better than what was before but does not cross the line @skynavga ?

andreastai: Thanks for the clarifying text, @palemieux. This will help the user to understand the issue.

I agree with @nigelmegitt that the text does not give the impression to be exclusive for Latin or Japanese text. To address @skynavga concern the following could be added:

> as is the case for Latin text **(or other left-to-right scripts)** ...as is the case for Japanese text **(or other right-to-left scripts)**

I do not think that the commit is "over-explaining" or that it adds "clutter". On the contrary: it provides the necessary examples for users and TTML experts. That this question needed to be discussed in the TTWG and that the issue was possibly overseen is an indication that also the later group could make good use of it.

andreastai (approved):

skynavga:
I would note that @palemieux, @nigelmegitt, and myself all approved the original PR commit at 2235c34969136e1d9b21cbed10d837fd49a8323b.

And that it was the introduction of new changes to address https://github.com/w3c/ttml1/pull/281#pullrequestreview-84006580 that we now have a disagreement.

spec/ttml1.xml
```diff
@@ -5070,6 +5070,9 @@ before edge, the second applies to the start and end edges, and the third applie
 to the after edge.
 If four <loc href="#style-value-length">&lt;length&gt;</loc> specifications are provided, then they apply to before, end,
 after, and start edges, respectively.</p>
+<note role="clarification">
+<p>Pecentage values are relative to the dimension of the region to which they apply. For example, if the before and after edges correspond to the top and bottom edges of the region, as is the case for Latin text where <att>tts:writingMode</att> is equal to <code>"lrtb"</code>, then percentage values that apply to either of the two edges are relative to the height of the region. Conversly, if the before and after edges correspond to the right and left edges of the region, as is the case for Japanese text where <att>tts:writingMode</att> is equal to <code>"tbrl"</code>, then percentage values that apply to either of the two edges are relative to the width of the region.</p>
```
nigelmegitt: just spotted a typo on re-reading: "conversly" -> "conversely".

nigelmegitt:
@skynavga the disagreement here is about how detailed and useful we make a clarifying note. The text is not normative, and at least I and @tairt feel it is better in this case to include more explanatory text. @tairt made a proposal to address your concern in https://github.com/w3c/ttml1/pull/281#discussion_r159221281 - perhaps you could respond to that?

skynavga (approved):
Think this is too cluttered, and could be more succinct, but will approve in the interest of moving forward.

PR #283, opened 2017-12-12
Deprecate offset-time and fraction component in smpte time base

### LINKED ISSUE #211: Use of offset time expression in smpte mode.
Need to address semantics of using an offset-time expression, e.g., 543t, 692.3f, etc, when time base is smpte continuous mode.

[1] https://www.w3.org/TR/ttml1/#time-expression-semantics-smpte

### LINKED ISSUE #213: Use of fractional seconds in clock-time expression in smpte mode.
Need to address semantics of using a clock-time expression that contains fractional seconds when time base is smpte continuous mode. As there is no current prohibition on the presence of the fractional component in clock-time expressions, the current semantics of the seconds variable of the countedFrames expression is not well-defined.

[1] http://w3c.github.io/ttml2/spec/ttml2.html#time-expression-semantics-smpte

See also https://github.com/w3c/ttml2/issues/165.

palemieux:
Close #211 and Close #213


skynavga (changes requested):
We cannot deprecate a feature in an edition of TTML1; only in a new version of TTML, namely TTML2.

palemieux:
Deprecation merely indicates future plans for removal, and does not forbid the feature, and thus can be enacted in TTML1.

skynavga:
I disagree, since it impacts what warnings a validator will emit by default. The best we can do in TTML1 is to say that a future version is expected to deprecate it and recommend it not be used, which must be placed in a note. I believe there is already such a warning note in the current TTML1 2nd Ed, so please follow that model.

palemieux:
Actually a warning is really appropriate since the deprecated forms should not be used since a behavior is not specified.

skynavga:
Maybe, but it would have to be optional. We can't deprecate a feature in an edition, only a new version, and for TTML, that means a new major version number, which is the only versioning we use for TTML (in contrast with IMSC which has gone another way).

palemieux:
> We can't deprecate a feature in am edition,

When creating an Edited Recommendation, the [W3C process](https://w3c.github.io/w3process) prevents the addition of features, presumably for IPR reasons, but does not prevent deprecation of features.

nigelmegitt:
@skynavga just for my understanding, are you asserting that TTML1 Third Edition falls into the class of "Edited Recommendation" according to the W3C Process?

nigelmegitt (approved):
I'm happy with these changes and don't see a problem deprecating formally syntax that had a poorly defined semantic anyway. 

We should separately raise an issue for styling of deprecated features to highlight them, assuming we go ahead with any deprecations - I understand from the conversation that consensus for this is not yet clear.

skynavga:
@nigelmegitt yes; we published 2nd Ed as an PER/ER [1], so the same applies for 3rd Ed

[1] https://www.w3.org/TR/2013/PER-ttaf1-dfxp-20130709/

nigelmegitt (approved):

skynavga (approved):

PR #284, opened 2017-12-12
Added fractional time semantics to Annex N.2

### LINKED ISSUE #188: Appendix N.2 omits fractional time expressions.
Appendix N.2 [1] describes the cases when time expressions include frames and subframes, and when it includes ticks, but not when it includes fractions of seconds. This usage is present in EBU-TT-D [2].

[1] http://www.w3.org/TR/ttml1/#time-expression-semantics-media
[2] https://tech.ebu.ch/docs/tech/tech3380.pdf

This could be fixed by adding the following alternative equation for M in the green box in this section immediately after the equation that uses frames and subFrames:

++++
OR

M = referenceBegin + 3600 \* hours + 60 \* minutes + seconds + fraction
++++

and changing

---
>> the hours, minutes, seconds, frames, subFrames components are extracted from time expression if present, or zero if not present;

to 

> the hours, minutes, seconds, frames, subFrames, fraction components are extracted from time expression if present, or zero if not present;

One could argue that this interpretation is obvious and does not require elaboration however I think it should be present for completeness. Otherwise when this appendix is made normative it could be interpreted that fractions of seconds are not permitted in time expressions when the timeBase is media, which I believe would be an error.

(raised by Nigel Megitt on 2014-05-02)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/306

palemieux:
Closes #188
Used prose from [TTML2](https://github.com/w3c/ttml2/commit/3fdda37d50fc504dec1863f82a5210dff95bd049)


nigelmegitt (approved):
Confirmed it matches the TTML2 modified language.

Separately, I wonder if in both places the language is ideal, since it introduces the concept of a real value seconds component in prose, without giving it a label. I worry that an implementer might miss the subtlety and discard the fraction component. Perhaps a clearer way to do this is to explicitly add `+ fraction` or to define a new `rSeconds` conceptual entity where `rSeconds = seconds + fraction`, to match the prose about the real valued seconds component.

Regardless of this observation, this pull request (and the existing one in TTML2) is an improvement.

skynavga (approved):

PR #287, opened 2017-12-13
Add note re: semantics of circular actor references

### LINKED ISSUE #201: Agent and actor circular references.
A ttm:agent can contain a ttm:actor. A ttm:actor can reference a ttm:agent. If an actor references its own parent agent then that would create a scenario that needs special handling in simple or naive implementations.

A major structural change to TTML1 is unwarranted at this stage however an editorial note to warn implementors of this would be helpful.

Alternatively an erratum prohibiting this scenario could be added - it doesn't have any obvious semantic meaning. Even if an actor plays himself, like Steve Coogan in The Trip playing Steve Coogan, the correct structure would be to have an agent whose type is "person", say xml:id="steveCooganTheActor" and an agent whose type is "character", say xml:id="steveCooganTheCharacter" that contains an actor whose agent="steveCooganTheActor".

(raised by Nigel Megitt on 2015-08-04)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/401

palemieux:
Close #201


spec/ttml1.xml
```diff
@@ -8581,6 +8581,13 @@ agent with another agent that portrays the character.</p>
 <p>The <att>agent</att> attribute of a <el>ttm:actor</el> element must
 reference a <emph>significant</emph> <el>ttm:agent</el> element that denotes the agent
 acting the part of a character.</p>
+<note role="clarification">
+<p>The semantics associated with an <el>ttm:actor</el> element referencing its parent 
+<el>ttm:agent</el> element are not defined. In the case of an actor playing themselves, 
+like Steve Coogan in The Trip playing Steve Coogan, two <el>ttm:agent</el> elements are used instead:
```
nigelmegitt: The "are used instead" language seems strange - I think this is describing good practice, so this clause could be rephrased:

_the expected practice is to use two `ttm:agent` elements instead:_

spec/ttml1.xml
```diff
@@ -8581,6 +8581,13 @@ agent with another agent that portrays the character.</p>
 <p>The <att>agent</att> attribute of a <el>ttm:actor</el> element must
 reference a <emph>significant</emph> <el>ttm:agent</el> element that denotes the agent
 acting the part of a character.</p>
+<note role="clarification">
+<p>The semantics associated with an <el>ttm:actor</el> element referencing its parent 
+<el>ttm:agent</el> element are not defined. In the case of an actor playing themselves, 
+like Steve Coogan in The Trip playing Steve Coogan, two <el>ttm:agent</el> elements are used instead: 
+the first element (corresponding to Steve Coogan the person) has <att>type</att> attribute equal to <code>person</code>, 
+while the second element (corresponding to Steve Coogan the character) has (i) a <att>type</att> attribute equal to <code>character</code> and (ii) a child <el>ttm:actor</el> element that references the first <el>ttm:agent</el> element.</p>
```
nigelmegitt: This wording is fine but doesn't address the need to warn implementers that special handling is required. Perhaps something like:

_Despite the lack of defined semantics, it is nevertheless syntactically permitted for a `ttm:actor` element to reference its parent `ttm:agent` element; implementers should take care to handle this situation, for example avoiding infinite loops etc._

nigelmegitt (changes requested):
Requesting stronger advice to implementers to address the issue.

palemieux:
@nigelmegitt The recommendation against circular references is now a _should not_ in light of:

- the original comment at #201
- the comment at https://github.com/w3c/ttml1/issues/201#issuecomment-351272720
- the resolution to issue #213 

The explanation and examples have been split into different notes.

xchange11 (approved):
Looks good to me.

nigelmegitt (approved):
Looks good to me. Thanks for making the changes.

PR #288, opened 2017-12-13
Clarify [associate region] algorithm

### LINKED ISSUE #194: Ambiguous definition for determination of descendant region identifier.
Step 3 of the [association region] procedure in Section 9.3.2 is ambiguous if multiple descendants are associated with a region (possibly distinct).

Suggest changing from:

"if the element contains a descendant element that specifies a region attribute, then the element is associated with the region referenced by that attribute;"

to

"if the element contains a descendant element that specifies a region attribute, then the element is associated with the first region referenced by that attribute using a breadth-first pre-traversal search of descendant elements;"

(raised by Glenn Adams on 2014-08-27)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/341

palemieux:
Close #194


spec/ttml1.xml
```diff
@@ -7454,27 +7454,25 @@ associated (owning) element node.</p>
 <gitem>
 <label>[associate region]</label>
 <def>
-<p>A <loc href="#element-vocab-type-content">Content</loc> element is associated with a region according
-to the following ordered rules, where the first rule satisfied
-is used and remaining rules are skipped:</p>
+<p>A <loc href="#element-vocab-type-content">Content</loc> element <emph>E</emph> is associated 
+with a region <emph>R</emph> if
+any of the following assertions is true:</p>
```
skynavga: Why are you change from first satisfied rule to any satisfied rule? The existing text orders the rules, which may lead to a different result when the rules are not ordered. This seems problematic and unjustified.

skynavga (changes requested):
Overall, this change is too extensive, and rewrites prose when not necessary, which may lead to different interpretations. Please restore the original text and only make the absolute minimum necessary normative changes.

palemieux:
@skynavga See discussion #194. You've had 12 months to review the proposed algorithm. I have addressed every comment you have raised against it.

spec/ttml1.xml
skynavga: Clearly this (new) algorithm is broken since it permits an element to be associated with four different regions, one for each rule. The original issue [1] pointed out that an ambiguity existed, which your rewrite has not fixed, but instead made worse (by increasing the ambiguity).

[1] https://github.com/w3c/ttml1/issues/194#issue-115107118

skynavga (changes requested):

nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml1/issue-194-associate-region-clarification-build/index.html

spec/ttml1.xml
nigelmegitt: @skynavga see https://github.com/w3c/ttml1/issues/194#issuecomment-272434373 in which the group agreed that content elements can be associated with multiple regions simultaneously.

If you don't believe this is true, look again at https://github.com/w3c/ttml1/issues/194#issuecomment-268853318 and consider that a `<div>` element, for example, may set styles that are important for all of its temporally active descendants, and that those descendants may each be selected into a different region. Therefore that `div` itself must be selected into multiple regions.

nigelmegitt (approved):
I think this works and satisfies the conclusion agreed by the group for #194.

I don't much like the "aren't" shortening in [construct intermediate document] step 2.d but that was not introduced by this pull request.

skynavga:
First, I would point out that the algorithm proposed by @palemieux in https://github.com/w3c/ttml1/issues/194#issuecomment-257047647 retained the principle of ordered rules with the first one winning. Next, I point out that your interpretation in https://github.com/w3c/ttml1/issues/194#issuecomment-268757592 shows that it is much more complex than saying any element can have multiple regions (a point not addressed in the proposed new algorithm). Next, I point out that the text proposed in this PR does not match what is described in https://github.com/w3c/ttml1/issues/194#issuecomment-272434373. And, finally, I had a placeholder for a review in https://github.com/w3c/ttml1/issues/194#issuecomment-272434373, and have not completed that process. Not only do I need to check what breakage this may cause in TTT, but I also need to check if this change violates a previous agreement by the WG to not target the same content to multiple regions. [It is one thing to say that some of the non-leaf ancestor hierarchy might be replicated in other regions, and another to say that leaf content elements, specifically, anonymous spans, may be targeted to multiple regions (in the same ISD).] As currently specified, the new changes seem to go far beyond the original issue posted, and may very will be a breaking, substantive change. I am not at all prepared to accept the current proposed text.

nigelmegitt:
@skynavga ok, but time is very limited to resolve this now, and it is important that we do so. The existing text is broken in the sense that the issue describes, and we must have new text agreed, preferably before we meet on Jan 9.

palemieux:
Updated to actually match https://github.com/w3c/ttml1/issues/194#issuecomment-257047647

spec/ttml1.xml
```diff
@@ -7469,27 +7469,25 @@ associated (owning) element node.</p>
 <gitem>
 <label>[associate region]</label>
 <def>
-<p>A <loc href="#element-vocab-type-content">Content</loc> element is associated with a region according
-to the following ordered rules, where the first rule satisfied
-is used and remaining rules are skipped:</p>
+<p>A <loc href="#element-vocab-type-content">Content</loc> element <emph>E</emph> is associated 
+with a region <emph>R</emph> if
+any of the following ordered assertions is true, where the first assertion satisfied is used and remaining assertions are skipped:</p>
```
nigelmegitt: First, I'd use "conditions" instead of "assertions" everywhere on this line for clarity.

Second, as this is written, I don't understand what difference it makes to say "any ... is true" and follow it by "where the first is used and remaining ones are skipped". Surely this is simply the logical consequence of an OR relationship - once you've found one that evaluates to true then you know the answer is true and you don't have to evaluate the others.

Hence I think the "where the first assertion satisfied is used and remaining assertions are skipped" text is unnecessary and should be removed.

It would be necessary if the mechanism by which the element was associated with the region were used later, but I believe it is not.

**However**, see comments below - it may be that the intent is not being clearly captured, and there is a useful stopping mechanism that can return an early "false".

spec/ttml1.xml
```diff
@@ -7469,27 +7469,25 @@ associated (owning) element node.</p>
...
-then the element is associated with the region referenced by the most
-immediate ancestor that specifies this attribute;</p></item>
-<item><p> if the element contains a descendant element that
-specifies a <loc
-href="#layout-attribute-region"><att>region</att></loc> attribute,
-then the element is associated with the region referenced by that
-attribute;</p></item>
+
+<item><p>if <emph>E</emph> specifies a <loc href="#layout-attribute-region">
+<att>region</att></loc> attribute that references <emph>R</emph>;</p></item>
```
nigelmegitt: My reading of this condition is that if E specifies a region attribute that references a region _other than_ R then this condition is skipped and the next is evaluated. However the introduction of the "stopping" rule at line 7474 makes me wonder if in that case the intent is to stop and return false, since the condition "E specifies a region attribute" is true, but the region attribute that it specifies is different. If that is the intent then this needs to be reworded. Either way it would be useful to clarify.

spec/ttml1.xml
```diff
@@ -7469,27 +7469,25 @@ associated (owning) element node.</p>
...
+
+<item><p>if the <loc href="#layout-attribute-region">
+<att>region</att></loc> attribute of the first ancestor element of <emph>E</emph> that specifies that attribute references <emph>R</emph>;</p></item>
```
nigelmegitt: Again, if the first ancestor element that specifies the attribute references a region other than R, should we return false immediately?

nigelmegitt: Is "first ancestor element" clear? "Closest ancestor" is perhaps less ambiguous, since in this case it is clear that a parent element is closer than a grandparent element. "First" implies some ordering, but that ordering is not specified. Possibly some readers could interpret it as being the most distant.

spec/ttml1.xml
```diff
@@ -7469,27 +7469,25 @@ associated (owning) element node.</p>
...
+
+
+<item><p> if a descendant element of <emph>E</emph> specifies a
```
nigelmegitt: Is this true for _any_ descendant element? What if the child specifies R' (where R' is a different region to R) but one of its children specifies R?

nigelmegitt (changes requested):
Rereading after the more recent change, I am no longer sure if this is right - changing my review to a "request changes" pending further discussion/enlightenment.

spec/ttml1.xml
```diff
@@ -7503,9 +7503,13 @@ to the following ordered rules, where the first rule satisfied
...
 <item><p>if the element specifies a <loc
-href="#layout-attribute-region"><att>region</att></loc> attribute,
-then the element is associated with the region referenced by that
-attribute;</p></item>
+href="#layout-attribute-region"><att>region</att></loc> attribute that
+references the region,
+then the element is associated with the region;</p></item>
+<item><p>if the element specifies a <loc
+href="#layout-attribute-region"><att>region</att></loc> attribute that does not
+references the region,
```
nigelmegitt: Typos:

s/references/reference

nigelmegitt (changes requested):
Just one typo, then I think this is a good clarifying change following today's f2f discussion.

nigelmegitt (commented):
Looking at the remainder of the bullets, the step that is now numbered 4 suffers from the identical problem. Would you be prepared to make the same modification @palemieux ?

palemieux:
@nigelmegitt An element can be associated with two regions if two of its descendents reference two different regions, right?

nigelmegitt:
@palemieux Yes. My point is just that the same structural change to the text applies to that later bullet.

If we thought that the meaning of "...then the element is associated with the region referenced by that
-attribute;..." was misleading in one bullet point then it is equally misleading in the other bullets in which it appears, and the structural change to the text made by this pull request should be applied to all those instances within this algorithm.

palemieux:
@nigelmegitt My understanding of the discussion is different: only step (1) was modified.

Step (4) cannot be modified as follows;

```
(4) if the element contains a descendant element that specifies a region attribute that references the region, then the element is associated with the region;
(5) if the element contains a descendant element that specifies a region attribute that does not reference the region, then the element is not associated with the region;
```

otherwise, none of the text in the following example would be shown since `<body>` would not be associated with either `r1` or `r2` and thus would be pruned.

```
 <body>
   <div>
     <p>
       <span region="r1">in r1</span>
       <span region="r2">in r2</span>
     </p>
   </div>
 </body>
```

[edit: typo]

skynavga (changes requested):
If the resulting changes do not produce the same results as the code at [1], then I cannot accept the PR without further and very careful consideration.

[1] https://github.com/skynav/ttt/blob/240ebb8d27a1a4b80f73a833dcc8bcb18214af9b/ttt-ttx/src/main/java/com/skynav/ttx/transformer/isd/ISD.java#L605-L631

nigelmegitt:
@palemieux OK but a similar change is needed for bullet 3:

> 3. if some ancestor of that element specifies a region attribute, then the element is associated with the region referenced by the most immediate ancestor that specifies this attribute;

should be:

_3. if some ancestor of that element specifies a region attribute, and the most immediate ancestor that specifies this attribute references the region, then the element is associated with the region;
4. if some ancestor of that element specifies a region attribute, and the most immediate ancestor that specifies this attribute does not reference the region, then the element is not associated with the region;_


The wording in bullet 4 (will become 5 if the above change to 3 is made) is still a bit wrong. It says: 

> 4. if the element contains a descendant element that specifies a region attribute, then the element is associated with the region referenced by that attribute;

However this does not deal with the case that multiple descendant elements may each specify different region attributes. I think it needs to be reworded to:

_4.  if the element contains a descendant element that specifies a region attribute **that references the region**, then the element is associated with the region;_

palemieux:
@skynavga Do you still agree with your recommendation at the end of https://github.com/w3c/ttml1/issues/194#issuecomment-224911805 ?

palemieux:
@nigelmegitt I am getting lost in the numbering, can you provide the full algorithm? Also can you confirm whether you agree with @skynavga's recommendation at https://github.com/w3c/ttml1/issues/194#issuecomment-224911805?

nigelmegitt:
@palemieux really?

Now in this pull request:

> [associate region]
>
>    A Content element is associated with a region according to the following ordered rules, where the first rule satisfied is used and remaining rules are skipped:
1  if the element specifies a region attribute that references the region, then the element is associated with the region;
2  if the element specifies a region attribute that does not reference the region, then the element is not associated with the region;
3  if some ancestor of that element specifies a region attribute, then the element is associated with the region referenced by the most immediate ancestor that specifies this attribute;
4  if the element contains a descendant element that specifies a region attribute, then the element is associated with the region referenced by that attribute;
5  if a default region was implied (due to the absence of any region element), then the element is associated with the default region;
6  the element is not associated with any region.

Proposal:

_1  if the element specifies a region attribute that references the region, then the element is associated with the region;
2  if the element specifies a region attribute that does not reference the region, then the element is not associated with the region;
3. if some ancestor of that element specifies a region attribute, and the most immediate ancestor that specifies this attribute references the region, then the element is associated with the region;
4. if some ancestor of that element specifies a region attribute, and the most immediate ancestor that specifies this attribute does not reference the region, then the element is not associated with the region;
5  if the element contains a descendant element that specifies a region attribute that references the region, then the element is associated with the region;
6  if a default region was implied (due to the absence of any region element), then the element is associated with the default region;
7  the element is not associated with any region._

In the proposed step 5 and the existing step 4, there is another potential issue in that it does not scope it to the first descendant level that specifies a region attribute. In other words, if we are querying for association with R1, and we get to step 5 in the proposed algorithm above, and a descendant element E2 references R2, but then E2 itself has a descendant E3 that references R1, then that step returns a false "true":

```
<div>
  <p region="R2">
    <span region="R1">...</span>
  </p>
</div>
```

The `div` is apparently associated with R1 according to the descendant element test.

nigelmegitt:
Closing as per resolution made in meeting today.

PR #290, opened 2017-12-14
State that LWSPs are permitted in tts:fontFamily

### LINKED ISSUE #248: Handling of LWSP in tts:fontFamily
See w3c/imsc#221 and w3c/ttml2#315 for background.

Whether or not LWSP is permitted around comma delimiters in `tts:fontFamily` is ambiguous. There is evidence that some documents include such spaces, whereas some processors may not accept them. For example one test case (fontFamily009.ttml) in the TTML1 test suite includes such spaces.

Proposed resolutions:

1. Prohibit LWSP in document instances but note that processors should be tolerant of them due to existing content that may include them.
2. Permit LWSP in document instances but note that some existing processors may not be tolerant of them.

palemieux:
Close #248


nigelmegitt (approved):
Looks good to me. 

For preference I think I'd have added this note after the existing note, but the order of notes is not really important here.

spec/ttml1.xml
```diff
@@ -4182,6 +4182,11 @@ as a style property only to those element types indicated in the following table
 </tbody>
 </table>
 <note role="clarification">
+<p>While permitted, the use of linear whitespace (LWSP) around the comma delimiters in the value of this 
+attribute is not recommended for maximal compatibilty with processor implementations.
```
skynavga: Is there a processor for which this is known to cause a problem?

skynavga (changes requested):

andreastai (approved):
Looks good to me. @skynavga Does it resolve your concern if "with processor implementation"? If not, what is your change proposal?

spec/ttml1.xml
skynavga: If we don't actually know of an incompatible processor, then please remove "... for maximal compatibility with processor implementations".

skynavga (changes requested):

palemieux:
@skynavga This PR reflects the consensus text at #248, which was the result on lengthy discussions which you were part of. Unless information has changed since the drafting of the text and/or you plan on filing a formal objection against the text of the PR, I do not see a compelling reason for the group to spend valuable time reopening this issue.

skynavga:
I do not accept this change as stated. If you can't point out an affected implementation, then it is mere speculation to suggest there is an incompatibility. While I accept the recommendation against use, I do not accept a motivation based on speculation.

palemieux:
@skynavga The question is whether you can still live with the consensus text at #248 , which would participated in. If not, i.e. you plan on formally objecting, the group will need to reopen the debate, which will undoubtedly take time away from closing other issues. What say you?

skynavga:
Feel free to put this on the WG agenda. I do not accept the current text.

nigelmegitt:
@skynavga you reacted with a thumbs up to https://github.com/w3c/ttml1/issues/248#issuecomment-304038399 to indicate your preference for the option:

> Permit LWSP in document instances but note that some existing processors may not be tolerant of them.

If you have a wording change that would achieves this option in a more satisfactory way to you, please propose it.

skynavga:
@nigelmegitt how many times shall I repeat myself?

please remove "... for maximal compatibility with processor implementations".

nigelmegitt:
@skynavga Right, but that's then incomplete, since removing that text means the pull request no longer addresses the consensus "but note that some existing processors may not be tolerant of them" - that's the bit that needs replacement wording please.

skynavga:
If nobody can point at such a processor, then that statement is factually wrong.

palemieux:
> If nobody can point at such a processor, then that statement is factually wrong.

These and other many other issues were taken into account while crafting the consensus text at #248. One aspect of the consensus cannot be reopened without reopening it completely. @skynavga What say you?

andreastai:
Could we add for example some text that says something along the line "...because of some ambiguity in TTML 2nd edition that could have led to a different interpretation" and than remove the contentious part "for maximal compatibilty with processor implementations"? Would this work for you @skynavga. 

I agree with @nigelmegitt that without an explaination why this is not recommended it would be incomplete. But I also think that if we re-word it and omit the "processor part" the informative value of the note is still there.

skynavga:
Firstly, I reject the notion that there has ever been an ambiguity here, or that an implementation that failed to study the relevant specifications and reached the wrong conclusion should be permitted to continue to exist.

Secondly, to my knowledge, nobody has actually pointed out the existence of such an implementation, which to me, means that the contention that a problem exists is merely conjecture.

My conclusion is that no change is necessary; however, I do not object to adding a note (for those readers who fail to do their research elsewise)  that makes the most minimum disrecommendation on the off-chance that such an implementation exists, but I am not willing to make reference to the speculated existence of such an implementation.

Please show me such an implementation, and I would be willing to change my mind.

skynavga (approved):
Approve with the changed text, and comment yet again that it is not a requirement, and in my opinion, poor practice to document rationale.

PR #291, opened 2017-12-14
Handle forward interoperability of attribute extensions in TT NS

### LINKED ISSUE #196: Handling forward interoperability of attribute extensions in TT namespaces.
Although there was an intent to permit forward interop in the presence of attribute extensions in TT namespaces (in future versions, like TTML2), this was not satisfactorily document in TTML1 Section 4 [1] step 3, which presently states:

"3. pruning all attribute information items having expanded names such that the namespace URI of the expanded names are not listed in Table 1 - Namespaces,"

In order to resolve, this needs to be changed to read:

"3. pruning all attribute information items having expanded names such that the namespace URI of the expanded names are not listed in Table 1 - Namespaces, or, if listed in Table 1 - Namespaces, are not members of the collection of attributes defined by the associated Abstract Document Type for use with the owning element information item,"

Note that step 1 already handles this case for element information items:

"1. pruning all element information items whose names are not members of the collection of element types defined by the associated Abstract Document Type, then"

[1] http://www.w3.org/TR/2013/REC-ttml1-20130924/#ttml-content-doctype

(raised by Glenn Adams on 2015-01-01)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/362

palemieux:
Closes #196

### DIFF
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -1075,7 +1075,9 @@ is defined as empty in the associated <emph>Abstract Document Type</emph>, and t
 <p>pruning all attribute
 information items having expanded names such that the namespace URI of the
 expanded names are not listed in <specref
-ref="namespace-vocab-table"/>,</p>
+ref="namespace-vocab-table"/>, or, if listed in <specref ref="namespace-vocab-table"/>, are not 
+members of the collection of attributes defined by the associated <emph>Abstract Document Type</emph> for use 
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
LGTM

skynavga (approved):

andreastai (approved):

PR #293, opened 2017-12-15
Specify application defaults for frame rate multiplier and sub-frame rate

### LINKED ISSUE #192: Use of application default frameRateMultipler, subFrameRate.
The defaulting rules for frameRate allow use of a default application specific value if frameRate is not specified, or 30 if no default application value applies. If a default application value is used, then a potential default application value should also be used with frameRateMultiplier and subFrameRate as well. However, the current defaulting language for these two parameters merely states that "1 1" and "1" be the default values respectively.

This should be fixed in an errata for TTML1.0SE and fixed inline in TTML2.

(raised by Glenn Adams on 2014-08-08)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/333

palemieux:
Closes #192

### DIFF
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -2461,8 +2461,8 @@ numerator | denominator
 </table>
 <p>A frame rate multiplier is used when the desired frame rate cannot
 be expressed as an integral number of frames per second.</p>
-<p>If not specified, the frame rate multiplier must be considered to be equal to one (1:1).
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
Looks good to me.

andreastai (approved):

skynavga (approved):

PR #295, opened 2017-12-15
Add <br> to list of elements to which style applies

### LINKED ISSUE #249: style namespace attributes in br
Reported by @cconcolato :

The "XML representation" for `<br>` element says it allows the style
attribute (see https://www.w3.org/TR/ttaf1-dfxp/#content-vocabulary-br),
but the style attribute definition says it is not allowed on `<br>`
(https://www.w3.org/TR/ttaf1-dfxp/#style-attribute-style).

palemieux:
Close #249

### DIFF
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -3497,6 +3497,7 @@ to reference one or more <el>style</el> elements each of which define a style (p
 element types:</p>
 <ulist>
 <item><p><loc href="#document-structure-vocabulary-body"><el>body</el></loc></p></item>
+<item><p><loc href="#content-vocabulary-br"><el>br</el></loc></p></item>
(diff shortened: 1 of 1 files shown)


nigelmegitt (approved):
Looks good to me.

andreastai (approved):
Looks good to me to make the spec consistent. But it is really confusing to give the option of styling to an element that can not be styled at all.

skynavga:
@tairt since we map ``<br/>`` to ``<fo:character/>``, and the latter admits certain style properties (two of which we prescribe), about which see https://www.w3.org/TR/xsl11/#prapply, then it is not unreasonable; however, it does raise the issue if we should specify that certain tts:* apply to ``br``, such as tts:direction and tts:unicodeBidi for example, where I can see how they would have semantic use.

skynavga (approved):

PR #296, opened 2017-12-16
Removed redundant clause at 8.4.1.3 Chained Referential Styling

### LINKED ISSUE #275: Chained Referential Styling spec vs example mismatch
While reading TTML1 I came on the following sentence in "8.4.1.3 Chained Referential Styling":
"If the same style property is specified in more than one referenced style set, then the last referenced style set applies, where the order of application starts from the affected element and proceeds to referenced style sets, and, in turn, to subsequent referenced style sets."

It talks about the "last" style applying. I wondered what "last" meant, according to which order. The next sentence says the order starts from the element to referenced style sets. This seems to me contradicting the example below in the spec.

I checked the rendering of IMSC.js and TTPE and both render the following code with: "Yellow Monospace" (in yellow) and "White Monospace" (in white).
```xml
<tt xml:lang="" xmlns="http://www.w3.org/ns/ttml">
<head>
<styling xmlns:tts="http://www.w3.org/ns/ttml#styling">
<style xml:id="s1" tts:color="white" tts:fontFamily="monospaceSerif"/>
<style xml:id="s2" style="s1" tts:color="yellow"/>
</styling>
</head>
<body>
<div>
<p style="s1">White Monospace</p>
<p style="s2">Yellow Monospace</p>
</div>
</body>
</tt>
```

This seems contradictory also to the algorithm specified in "8.4.4.2 Specified Style Set Processing", which says (removing unnecessary parts):
```
1. [initialization] initialize the specified style set SSS of E to the empty set;
2. [referential and chained referential styling] for each style element SREF referenced by a style attribute specified on E, and in the order specified in the style attribute, then, if SREF is a descendant of a styling element, merge the specified style set of SREF, SSS(SREF), into the specified style set of E, SSS(E);
4. [inline styling] for each style property P expressed as a specified styling attribute of E, merge P into the specified style set of E, SSS(E);
```

I interpret this as follows:
```
1. SSS(<p style="s2">) = empty
2. get the SSS(s2), i.e:
  1. SSS(s2) = empty
  2. get SSS(s1), i.e.:
     1. SSS(s1) = empty
     4. SSS(s1) = color:white
     merge SSS(s1) into SSS(s2): SSS(s2) = color:white
  4. merge inline property: SSS(s2) = color:yellow
  merge SSS(s2) into SSS(<p style="s2">):  SSS(<p style="s2">) = color:yellow
```

Conclusion: I think the example is correct. The sentence in 8.4.1.3 is wrong. Given that it is duplicating the algorithm of 8.4.4.2, I would suggest removing the sentence in 8.1.4.3. There is no need to duplicate text.

palemieux:
Closes #275


spec/ttml1.xml
```diff
@@ -6654,9 +6654,6 @@ ref="semantics-style-inheritance-content"/> below.</p>
 other out-of-line style properties, thus creating a chain of references starting at the
 affected element. When expressed in this manner, the association of style information
 is referred to as <emph>chained referential styling</emph>.</p>
-<p>If the same style property is specified in more than one referenced
```
skynavga, line 6685: I am opposed to removing this. If there is a problem with it, then let's fix it.

skynavga (changes requested):

nigelmegitt (dismissed):
This is fixing the wrong thing in my opinion - it is the example that is misleading not the sentence being removed here.

palemieux:
@nigelmegitt @skynavga Please propose a fix to the example then.

nigelmegitt:
@palemieux @skynavga I don't want to have this discussion in two places at once - right now it seems like it is happening in #275 so I'll wait for something closer to a conclusion there before proposing something additional here. I did propose an alternative example already in that thread.

skynavga:
I can accept removing the second paragraph of 8.4.1.3 and adding a note under the example as follows:

>__Note:__ The text of the second paragraph in this example is yellow, since ``tts:color='yellow'`` effectively overwrites (is merged over) the ``tts:color='white'`` that style ``s2`` obtains by a reference to style ``s1``.

palemieux:
Rendered at https://rawgit.com/w3c/ttml1/issue-275-redundance-clause-at-8-4-1-3-build/index.html

skynavga (approved):

andreastai (approved):

spec/ttml1.xml
```diff
@@ -6654,9 +6654,6 @@ ref="semantics-style-inheritance-content"/> below.</p>
...
-style set, then the last referenced style set applies, where the order of application starts from
-the affected element and proceeds to referenced style sets, and, in turn, to subsequent referenced style sets.</p>
```
nigelmegitt, line 6687: I would still like to add a replacement sentence that explains that when chained referential styling is used each referenced style is fully resolved before the reference is followed.

nigelmegitt (commented):
I don't believe this fully addresses the issue since I would like a clear short sentence to be added in place of the removed paragraph to explain that each referenced style must be resolved into specified styles before being followed.

PR #297, opened 2017-12-16
Clarify semantics of default overflow semantics implied by XSL-FO

### LINKED ISSUE #239: tts:overflow does not apply to the region area
`tts:overflow` states _the tts:overflow attribute is used to specify a style property that defines whether a region area is clipped or not if the descendant areas of the region overflow **its extent**_.

Presumably `tts:overflow` applies to the area of the region as optionally inset by `tts:padding`, and not the extent of the region.

Proposed solution: define a term for the inset area of a region and apply `tts:overflow` to it.

See https://github.com/w3c/ttml1/pull/233/commits/e8a7bdb034d26102b4358a0808dfaf96d2fa95aa#r113619948

skynavga:
Closes #239.

### DIFF
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -7682,6 +7682,10 @@ to satisfy alignment in the block progression dimension. For example, this assur
 the background color associated with the <el>body</el> element, if not <code>transparent</code>,
  will fill the containing region wholly.</p>
 </note>
+<note role="clarification">
+<p>For the purpose of performing presentation processing, the <code>auto</code> value of
+the XSL FO <code>overflow</code> property is interpreted as <code>visible</code>.</p>
+</note>
 </div3>
 <div3 id="semantics-region-layout-example">
 <head>Elaborated Example (Informative)</head>


palemieux (approved):
Do not merge until @tairt reviews

spec/ttml1.xml
```diff
@@ -7682,6 +7682,10 @@ to satisfy alignment in the block progression dimension. For example, this assur
 the background color associated with the <el>body</el> element, if not <code>transparent</code>,
  will fill the containing region wholly.</p>
 </note>
+<note role="clarification">
+<p>For the purpose of performing presentation processing, the <code>auto</code> value of
+the XSL FO <code>overflow</code> property is interpreted as <code>visible</code>.</p>
+</note>
 </div3>
```
andreastai, line 7689: Text should be added why this clarification is needed e.g.

"The overflow property of a fo:block element that have been mapped from body, div or p has the initial value of auto."

andreastai (changes requested):
See comments in code.

spec/ttml1.xml
skynavga: Really? In general, we don't provide motivation or rationale information, which is more appropriate in a user guide than a technical specification. Furthermore, here, we (have to) assume that the reader is familiar with XSL-FO. Further still, and more importantly, it is not true that "the overflow property of a fo:block element" [has been] "mapped from body, div, or p", since tts:overflow does not apply to the latter.

andreastai: > In general, we don't provide motivation or rationale information, which is more appropriate in a user guide than a technical specification.

As our discussion has shown clarification is more than needed.

> Furthermore, here, we (have to) assume that the reader is familiar with XSL-FO. 

Unfortunately we can not assume that the reader will be familiar with XSL-FO to that level of detail.

> Further still, and more importantly, it is not true that "the overflow property of a fo:block element" [has been] "mapped from body, div, or p", since tts:overflow does not apply to the latter.

Ok, than we can re-word it. It is not the overflow property that is mapped but body, div and p are mappped to an fo:block element. This fo:block element has an initial value for the overflow property. 

Proposal:

 "The fo:block element that have been mapped from body, div or p has an overflow property with the initial value of auto."

skynavga: Upon review of XSL-FO 1.1, Section B.4, we see that overflow does not apply to fo:block or fo:inline.

> Property overflow applies to: fo:region-body, fo:region-before, fo:region-after, fo:region-start, fo:region-end, fo:block-container, fo:external-graphic, fo:instream-foreign-object, and fo:inline-container.

andreastai:
In this case this PR does not need to be merged as the value "auto" will never be applied. #239 could be changed without any change.

nigelmegitt:
Closing as group agrees that this change is not beneficial.

Issue #217, opened 2016-10-30
[Specified Style Set Processing] Initial values only applied to regions


palemieux:
The _Specified Style Set Processing_ algorithm only mentions initial values _if the element type of E is the layout element type region, then set P′ to the initial value of property P_.

How do initial values of non-inheritable styling attributes ever apply to elements other than regions?

Shouldn't initial values be applied to all elements and not only region elements?

skynavga:
Duplicate of #200.

PR #298, opened 2017-12-18
Add initial value fallback to non-inheritable properties

### LINKED ISSUE #200: Add errata for initial value fallback language.
Add errata to back-port new language from TTML2 10.4.4.2 step (6) that deals with initial value fallback that was accidentally omitted from TTML1 8.4.4.2 step (6).

(raised by Glenn Adams on 2015-07-16)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/399

palemieux:
Closes #200 

See also https://github.com/w3c/ttml1/issues/217

### DIFF
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -7078,12 +7078,11 @@ following ordered sub-steps:</p>
 <item><p>if <emph>P</emph> is present in the specified
 style set of <emph>E</emph>, <emph>SSS(E)</emph>, then continue to the
 next style property;</p></item>
-<item><p>if <emph>P</emph> is inheritable and <emph>E</emph> is a <el>region</el> element
-and, or if <emph>P</emph> is not inheritable and applies to <emph>E</emph>, then set <emph>P&prime;</emph> to the initial value of
+<item><p>if <emph>P</emph> is inheritable and <emph>E</emph> is a <el>region</el> element, or if <emph>P</emph> is not inheritable and applies to <emph>E</emph>, then set <emph>P&prime;</emph> to the initial value of
 property <emph>P</emph>, where the initial value of a property is
 determined according to the specific property definition found above
 in <specref ref="styling-attribute-vocabulary"/>;</p></item>
-<item><p>if if <emph>P</emph> is inheritable and the element type of <emph>E</emph> is a <loc href="#element-vocab-type-content">Content</loc> element
+<item><p>if <emph>P</emph> is inheritable and the element type of <emph>E</emph> is a <loc href="#element-vocab-type-content">Content</loc> element
 type or anonymous span, then set <emph>P&prime;</emph> to the result of
 looking up the value of <emph>P</emph> in the computed style set of
 the immediate ancestor element of <emph>E</emph>, i.e.,
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -7070,20 +7070,20 @@ set of <emph>E</emph>, <emph>SSS(E)</emph>;</p></item>
 <item>
 <p><phrase role="strong">[implicit inheritance]</phrase> if the
 element type of <emph>E</emph> is not the styling element type
-<el>style</el>, then for each inheritable style property
-<emph>P<sub>INH</sub></emph> in the set of style properties defined
+<el>style</el>, then for each style property
+<emph>P</emph> in the set of style properties defined
 above in <specref ref="styling-attribute-vocabulary"/>, perform the
 following ordered sub-steps:</p>
 <olist>
-<item><p>if <emph>P<sub>INH</sub></emph> is present in the specified
+<item><p>if <emph>P</emph> is present in the specified
 style set of <emph>E</emph>, <emph>SSS(E)</emph>, then continue to the
-next inheritable style property;</p></item>
-<item><p>if the element type of <emph>E</emph> is the layout element
-type <el>region</el>, then set <emph>P&prime;</emph> to the initial value of
+next style property;</p></item>
+<item><p>if <emph>P</emph> is inheritable and <emph>E</emph> is a <el>region</el> element
+and, or if <emph>P</emph> is not inheritable and applies to <emph>E</emph>, then set <emph>P&prime;</emph> to the initial value of
 property <emph>P</emph>, where the initial value of a property is
 determined according to the specific property definition found above
 in <specref ref="styling-attribute-vocabulary"/>;</p></item>
-<item><p>if the element type of <emph>E</emph> is a <loc href="#element-vocab-type-content">Content</loc> element
+<item><p>if if <emph>P</emph> is inheritable and the element type of <emph>E</emph> is a <loc href="#element-vocab-type-content">Content</loc> element
 type or anonymous span, then set <emph>P&prime;</emph> to the result of
 looking up the value of <emph>P</emph> in the computed style set of
 the immediate ancestor element of <emph>E</emph>, i.e.,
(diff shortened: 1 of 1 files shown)


spec/ttml1.xml
```diff
@@ -7070,20 +7070,20 @@ set of <emph>E</emph>, <emph>SSS(E)</emph>;</p></item>
...
 <olist>
-<item><p>if <emph>P<sub>INH</sub></emph> is present in the specified
+<item><p>if <emph>P</emph> is present in the specified
 style set of <emph>E</emph>, <emph>SSS(E)</emph>, then continue to the
-next inheritable style property;</p></item>
-<item><p>if the element type of <emph>E</emph> is the layout element
-type <el>region</el>, then set <emph>P&prime;</emph> to the initial value of
+next style property;</p></item>
+<item><p>if <emph>P</emph> is inheritable and <emph>E</emph> is a <el>region</el> element
+and, or if <emph>P</emph> is not inheritable and applies to <emph>E</emph>, then set <emph>P&prime;</emph> to the initial value of
```
nigelmegitt: seems to be an unnecessary "and" at the beginning of this line.

spec/ttml1.xml
```diff
@@ -7070,20 +7070,20 @@ set of <emph>E</emph>, <emph>SSS(E)</emph>;</p></item>
...
 property <emph>P</emph>, where the initial value of a property is
 determined according to the specific property definition found above
 in <specref ref="styling-attribute-vocabulary"/>;</p></item>
-<item><p>if the element type of <emph>E</emph> is a <loc href="#element-vocab-type-content">Content</loc> element
+<item><p>if if <emph>P</emph> is inheritable and the element type of <emph>E</emph> is a <loc href="#element-vocab-type-content">Content</loc> element
```
nigelmegitt: typo "if if" -> "if"

spec/ttml1.xml
nigelmegitt: Some properties do not have an initial value specified, and that's not a spec error.

spec/ttml1.xml
```diff
@@ -7070,20 +7070,20 @@ set of <emph>E</emph>, <emph>SSS(E)</emph>;</p></item>
 <item>
 <p><phrase role="strong">[implicit inheritance]</phrase> if the
```
nigelmegitt, line 7086: I would prefer as a matter of style not to mix the setting of initial values of non-inheritable style properties into the implicit inheritance steps. Rather, I think the resolution to this issue would be better as a new step 7, such as [apply initial property values] along the lines of:

_7. **[apply initial non-inheritable style property values]** if the element type of E is not the styling element type `style`, then for each non-inheritable style property P in the set of style properties defined above in 8.2 Styling Attribute Vocabulary, perform the following ordered sub-steps:
     a. if P is present in the specified style set of E, SSS(E), then continue to the next style property;
     b. if P applies to E, then set P′ to the initial value of property P, where the initial value of a property is determined according to the specific property definition found above in 8.2 Styling Attribute Vocabulary, and is undefined if no initial value is defined;
     c. if the value of P′ is not undefined, then merge P′ into the specified style set of E, SSS(E)._

nigelmegitt (changes requested):
A couple of typos with this pull request as is, however I have proposed an alternative approach to resolving the issues that I think is clearer.

spec/ttml1.xml
palemieux: @nigelmegitt This would not match TTML2 ED. I think we should keep this structure, and file a separate ticket to consider updating both TTML2 and TTML1 to match your suggested structure.

nigelmegitt: OK I hadn't spotted that @palemieux , happy to go with your suggestion. Thanks for the pointer.

spec/ttml1.xml
```diff
@@ -7070,20 +7070,19 @@ set of <emph>E</emph>, <emph>SSS(E)</emph>;</p></item>
...
+<item><p>if <emph>P</emph> is present in the specified
 style set of <emph>E</emph>, <emph>SSS(E)</emph>, then continue to the
-next inheritable style property;</p></item>
-<item><p>if the element type of <emph>E</emph> is the layout element
-type <el>region</el>, then set <emph>P&prime;</emph> to the initial value of
+next style property;</p></item>
+<item><p>if <emph>P</emph> is inheritable and <emph>E</emph> is a <el>region</el> element, or if <emph>P</emph> is not inheritable and applies to <emph>E</emph>, then set <emph>P&prime;</emph> to the initial value of
 property <emph>P</emph>, where the initial value of a property is
 determined according to the specific property definition found above
 in <specref ref="styling-attribute-vocabulary"/>;</p></item>
```
nigelmegitt: Even if not using my suggestion at https://github.com/w3c/ttml1/pull/298/files#r157463436 right now, I think it would still be useful to note that if an initial value is absent from the property definition, then the specified value falls into the undefined class.

palemieux: Do you mean replace _then set P′ to the initial value of property P,_ with _then set P′ to the initial value of property P if one exists,_ ?

nigelmegitt: That would work but it would be a bigger change, since it would also mean that the step "if the value of P′ is not undefined, then merge P′ into the specified style set of E, SSS(E)." would need to be changed as well.

The smaller (and therefore usually preferable) change is simply to add something like the following to the end of the line this comment thread begins on:

_(for properties that do not define an initial value, the value of P′ is undefined)_

palemieux: @nigelmegitt if no initial value of property P exist than P' is not set, therefore P' remains undefined.... or is there something I missed?

nigelmegitt: @palemieux The point I'm trying to make is that the text directs the reader to the initial value of the property but for some properties _no initial value is defined_. In that case, what to do? I think it would be a helpful addition to clarify a) that such cases exist deliberately and b) that when they arise, P' remains undefined.

palemieux: @nigelmegitt _then set P′ to the initial value of property P if one exists_ seems to cover the case where no initial value is defined, since the clause would not be applied if no initial value exists, which would leave P' undefined... or am I missing a corner case... or is it a personal style issue?

nigelmegitt: @palemieux Maybe I've misunderstood, but I thought that if you do that ["if one exists"] then SSS(E) would not contain P', whereas if you adopt my approach SSS(E) would contain P', with P' set to an "undefined" class. So it seemed like a subtle but substantive difference.

palemieux: @nigelmegitt There is no "undefined" class in SSS(E). AFAIK "undefined" refers to the variable P' never having been set.

nigelmegitt: @palemieux OK, I see what you mean. If P' is undefined then it never gets added to SSS(E). So we should say:

_set P′ to the initial value of property P if such an initial value is defined_

I'm wary of saying "if one is defined" in case of ambiguity about what the "one" refers to.

palemieux: @nigelmegitt I have updated the PR per https://github.com/w3c/ttml1/pull/298#discussion_r157547379

nigelmegitt (approved):
Thanks, looks good to me.

A minor nit would be that we've ended up with two "ifs" rather than simply an additional "and an initial value is defined for P" clause. I think that may have been partially my fault...

Still, this is better than what was there before.

andreastai (approved):

skynavga (approved):
Good job handling this complex change. Further, upon reviewing this change, it suggests that I revert the order of steps 6 (b) and (c) in TTML2 10.5.4.2 to use the same ordering as was in TTML1.

PR #301, opened 2017-12-19
Clarify tts:fontSize semantics

### LINKED ISSUE #206: Use of 'em' units in tts:fontSize on region element is not well defined.

### LINKED ISSUE #214: Clarify the semantics of tts:fontSize="2em".
XSL, which TTML uses as a reference for 'em', states that "a relative length is a unit-based value that is measured against the current value of the font-size property" and "when the "font-size" property is
evaluated, the current font-size for use in evaluation is the font-size of the parent element"

While TTML does not specify a definition for the _current_ value of tts:fontSize, using the computed value of tts:fontSize of the parent would be consistent with XSL and the definition of the percentage metric in TTML.

See also #206

### LINKED ISSUE #215: Two value percentage fontSize not fully defined.
See w3c/ttml2#200

palemieux:
Clarify tts:fontSize specified in 'em' units (Closes #214 and closes #206)
Clarify tts:fontSize when specified using two % values (Closes #215)

This PR is intended to replace PR #279

The PR is based on the prose found in TTML2


palemieux:
See specification at https://rawgit.com/w3c/ttml1/issue-215-214-279-fontsize-semantics-build/index.html

andreastai (approved):
Looks good to me. Thank you! The examples in the informative notes make it very clear.

spec/ttml1.xml
```diff
@@ -4277,20 +4277,20 @@ as a style property only to those element types indicated in the following table
 </tr>
 <tr>
 <td><emph>Percentages:</emph></td>
-<td>if not <loc href="#content-vocabulary-span"><el>region</el></loc> element,
-then relative to parent element's font size; otherwise, relative to the
-<emph>Computed Cell Size</emph></td>
+<td>see discussion below</td>
```
nigelmegitt: Editorial point: the word "discussion" seems inappropriate for normative requirements. Suggest changing to "details" or even omitting altogether in favour of simple "see below".

spec/ttml1.xml
```diff
@@ -4277,20 +4277,20 @@ as a style property only to those element types indicated in the following table
...
 </tr>
 <tr>
 <td><emph>Animatable:</emph></td>
 <td>discrete</td>
 </tr>
 </tbody>
 </table>
+<p>The computed value of the property associated with this attribute consists of a 2-tuple
+the entries of which denote the computed values of the width and height components of the font size respectively.</p>
```
nigelmegitt, line 4354: Note for posterity: the rationale for this is described in https://github.com/w3c/ttml2/issues/200#issuecomment-305547161

spec/ttml1.xml
```diff
@@ -4277,20 +4277,20 @@ as a style property only to those element types indicated in the following table
...
 <p>If a single <loc href="#style-value-length">&lt;length&gt;</loc> value is specified, then this length applies
-equally to horizontal and vertical scaling of a glyph's EM square; if two
+equally to horizontal and vertical size of a glyph's EM square; if two
 <loc href="#style-value-length">&lt;length&gt;</loc> values are specified, then the first expresses the horizontal
-scaling and the second expresses vertical scaling.</p>
+size and the second expresses vertical size.</p>
```
nigelmegitt: Same comment as https://github.com/w3c/ttml2/pull/336/files#r157897318 :

This loses the sense that the font-defined glyph generally needs scaling to the precise display dimensions of the em square, and specifically always needs scaling anamorphically to achieve a different width to the height. I think this needs to be brought back in somehow, though I like the idea of specifying the resultant size of the em square unambiguously.

spec/ttml1.xml
```diff
@@ -4300,29 +4300,55 @@ one column in width.</p>
...
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
+
+<p>Relative <loc href="#style-value-length">&lt;length&gt;</loc> values
+that appear in this attribute, i.e., values expressed in percentage (<code>%</code>), cell (<code>c</code>), or EM (<code>em</code>) units,
+are resolved in relation to a referenced 2-dimensional size value consisting of two components, a width component and
+a height component.</p>
+
+
+<p>When a single relative <loc href="#style-value-length">&lt;length&gt;</loc> value is specified, then
+this <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the height component of the referenced value;
```
nigelmegitt: I like the idea of defining it in this way. It needs one further concept to be explicitly stated: if a single length value is specified, the resolved value relative to the reference is used for both the width and the height component of the computed value.

This concept is what is intended by the text on line 4290, "this length applies equally to horizontal and vertical ..." but that text is somewhat inadequate in my view: it is unclear that _applies_ means the same as the intended meaning _specifies_. Perhaps I should raise that as a separate issue, but I'd much rather have one pull request that sorts this tortuous font size issue out once and for all.

spec/ttml1.xml
```diff
@@ -4277,20 +4277,20 @@ as a style property only to those element types indicated in the following table
...
 </tr>
 <tr>
 <td><emph>Animatable:</emph></td>
 <td>discrete</td>
 </tr>
 </tbody>
 </table>
+<p>The computed value of the property associated with this attribute consists of a 2-tuple
+the entries of which denote the computed values of the width and height components of the font size respectively.</p>
 <p>If a single <loc href="#style-value-length">&lt;length&gt;</loc> value is specified, then this length applies
```
nigelmegitt, line 4355: See comment below - the word "applies" seems insufficient.

spec/ttml1.xml
```diff
@@ -4300,29 +4300,55 @@ one column in width.</p>
...
+the first <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the width component of the referenced value and
+the second <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the height component of the referenced value.</p>
+
+<p>If the relative unit is a percentage (<code>%</code>), then the referenced value is the <emph>Computed Cell Size</emph> if specified on the <loc href="#layout-vocabulary-region"><el>region</el></loc> element; or the parent element's computed font size otherwise.</p>
+
+<p>If the relative unit is cell (<code>c</code>), then the referenced value is the
+<emph>Computed Cell Size</emph>.</p>
+
+<note role="example">
+<p>For example, consider a paragraph (<code>p</code>) element <code>P</code>.
```
nigelmegitt, line 4385: As in the identical example in w3c/ttml2#336, the text is inconsistent with the span example on line 4339 - it seems odd that paragraph and span elements are not described in the same way. See also https://github.com/w3c/ttml2/pull/336/files#r157896644

spec/ttml1.xml
```diff
@@ -4300,29 +4300,55 @@ one column in width.</p>
...
 each character child of a <el>p</el> element is considered to be enclosed in an anonymous
 span.</p>
 <p>If a computed value of the property associated with this attribute is not supported,
 then a presentation processor must use the closest supported value.</p>
-<note role="elaboration">
-<p>In this context, the phrase <emph>closest supported value</emph> means the value for which the Euclidean distance between
-the computed font size and the supported font size is minimized. If there are multiple closest supported values equally distant from
-the computed value, then the value most distant from 0 (single length specification) or [0,0] (two length specifications) is used,
-i.e., the largest font size, is used.</p>
-</note>
```
nigelmegitt: Why drop this note? If I remember correctly it was added after feedback from @palemieux that implementers had difficulty understanding the correct algorithm for calculating the "closes supported value".

spec/ttml1.xml
```diff
@@ -4300,29 +4300,55 @@ one column in width.</p>
...
+the second <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the height component of the referenced value.</p>
+
+<p>If the relative unit is a percentage (<code>%</code>), then the referenced value is the <emph>Computed Cell Size</emph> if specified on the <loc href="#layout-vocabulary-region"><el>region</el></loc> element; or the parent element's computed font size otherwise.</p>
+
+<p>If the relative unit is cell (<code>c</code>), then the referenced value is the
+<emph>Computed Cell Size</emph>.</p>
+
+<note role="example">
+<p>For example, consider a paragraph (<code>p</code>) element <code>P</code>.
+If the <emph>Computed Cell Size</emph> is (24px,36px), and if
```
nigelmegitt, line 4386: Same comment as https://github.com/w3c/ttml2/pull/336/files#r157899921 :

As per the discussion in https://github.com/w3c/ttml2/issues/200#issuecomment-305547161 the sizes here are in logical pixels - in fact they are dimensionless until multiplied by the size of the pixel unit, whatever that turns out to be. Consider adding a parenthetical _(in logical pixels)_ to avoid confusion with "real" pixels.

spec/ttml1.xml
```diff
@@ -4300,29 +4300,55 @@ one column in width.</p>
...
+a height component.</p>
+
+
+<p>When a single relative <loc href="#style-value-length">&lt;length&gt;</loc> value is specified, then
+this <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the height component of the referenced value;
+when two relative
+<loc href="#style-value-length">&lt;length&gt;</loc> values are specified,
+the first <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the width component of the referenced value and
+the second <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the height component of the referenced value.</p>
+
```
nigelmegitt, line 4380: Seems to be missing the useful note as in https://github.com/w3c/ttml2/pull/336/files#diff-5ac39ab39265717d36dfd789d254ce1aR9801 about the use of anamorphic scaling.

nigelmegitt (changes requested):
Generally this is a good change headed in the right direction. Takes a lot of care to review in association with w3c/ttml2#336 - it is probably most efficient use of the group's time to consider both these pull requests together and, when happy with both, merge both. 

I've made some comments on both pull requests.

I've also requested some changes specific to this one, mainly editorial.

spec/ttml1.xml
nigelmegitt: Proposal:

_Note: Presentation processors are expected to scale glyphs so that the height of the em square as defined in the font resource matches the computed font size height, using an appropriate algorithm for the font. If the computed font size width differs from the computed font size height presentation processors additionally need to scale the em square anamorphically by the ratio of the width to the height._

spec/ttml1.xml
skynavga: normally, we say "see prose" which is consistent with XSL-FO and CSS usage

spec/ttml1.xml
```diff
@@ -4300,29 +4300,55 @@ one column in width.</p>
...
+<p>If the relative unit is a percentage (<code>%</code>), then the referenced value is the <emph>Computed Cell Size</emph> if specified on the <loc href="#layout-vocabulary-region"><el>region</el></loc> element; or the parent element's computed font size otherwise.</p>
```
skynavga: If we are trying to make this close to TTML2 changes, then this should go back into the table.

skynavga (changes requested):
My overall problem with this PR is that it attempts to incorporate some of the new TTML2 text, but does so only partly, and makes what appear to be arbitrary changes in that text as well. I suggest we focus on (1) identifying what is broken in TTML1 2ed, and (2) fix only that text using the absolutely minimum changes.

spec/ttml1.xml
nigelmegitt: Action on me to file this against TTML2 for discussion there.

spec/ttml1.xml
nigelmegitt: Needs filing against TTML2 also.

spec/ttml1.xml
palemieux: Added.

nigelmegitt (approved):

spec/ttml1.xml
nigelmegitt: Done as #548.

spec/ttml1.xml
nigelmegitt: Filed as #547

palemieux:
@skynavga Please review the revised PR

spec/ttml1.xml
```diff
@@ -4298,24 +4298,22 @@ as a style property only to those element types indicated in the following table
 </tr>
 <tr>
 <td><emph>Percentages:</emph></td>
-<td>if not <loc href="#content-vocabulary-span"><el>region</el></loc> element,
-then relative to parent element's font size; otherwise, relative to the
-<emph>Computed Cell Size</emph></td>
+<td>if specified on the <loc href="#layout-vocabulary-region"><el>region</el></loc> element, then relative to the <emph>Computed Cell Size</emph>; otherwise, relative to the parent element's computed font size.</td>
```
skynavga: please rewrite to use same structure as TTML2 text; namely, it should read:

if not region element, then relative to nearest ancestor styled element's computed font size; otherwise, relative to the Computed Cell Size

spec/ttml1.xml
```diff
@@ -4298,24 +4298,22 @@ as a style property only to those element types indicated in the following table
...
 </tr>
 <tr>
 <td><emph>Animatable:</emph></td>
 <td>discrete</td>
 </tr>
 </tbody>
 </table>
+<p>The computed value of the property associated with this attribute consists of a 2-tuple
+the entries of which denote the computed values of the width and height components of the font size respectively.</p>
```
skynavga, line 4354: I wonder if this new paragraph should go here, since "computed value" only appears in the notes farther down and in the paragraph about supported values. It may be better to move this to just prior to the paragraph that reads: "If the relative unit is cell (c), then the referenced value is the Computed Cell Size." However, I can see how putting it here may give guidance to interpreting the 5 paragraphs that precede this. Since I don't have a strong opinion, I can let this stand.

spec/ttml1.xml
```diff
@@ -4298,24 +4298,22 @@ as a style property only to those element types indicated in the following table
...
-equally to horizontal and vertical scaling of a glyph's EM square; if two
+equally to horizontal and vertical size of a glyph's EM square; if two
 <loc href="#style-value-length">&lt;length&gt;</loc> values are specified, then the first expresses the horizontal
-scaling and the second expresses vertical scaling.</p>
-<note role="example">
-<p>Use of independent horizontal and vertical font sizes is expected to be used
-with cell based units in order to denote fonts that are two rows in height and
-one column in width.</p>
+size and the second expresses vertical size.</p>
+<note role="elaboration">
```
skynavga: This note should be placed in the same relative position as in TTML2, i.e., after the paragraph starting with "When a single relative <length> value is specified...".

spec/ttml1.xml
```diff
@@ -4324,26 +4322,57 @@ glyphs do not have an EM square that is distinct from their font's EM square.</p
 <p>If horizontal and vertical sizes are expressed independently, then the
 units of the <loc href="#style-value-length">&lt;length&gt;</loc> values must be the same.</p>
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
+
+<p>Relative <loc href="#style-value-length">&lt;length&gt;</loc> values
+that appear in this attribute, i.e., values expressed in percentage (<code>%</code>), cell (<code>c</code>), or EM (<code>em</code>) units,
+are resolved in relation to a referenced 2-dimensional size value consisting of two components, a width component and
+a height component.</p>
```
skynavga, line 4372: Missing the following text present in TTML2:

>If the relative unit is a percentage, then the referenced value is either the computed value of the font size of the nearest styled ancestor element or the computed cell size. If the relative unit is cells (c), then the referenced value is the computed cell size. If the relative unit is EMs (em), then the referenced value is determined as if percentage units were used, where 1em equals 100%.

spec/ttml1.xml
```diff
@@ -4324,26 +4322,57 @@ glyphs do not have an EM square that is distinct from their font's EM square.</p
 <p>If horizontal and vertical sizes are expressed independently, then the
 units of the <loc href="#style-value-length">&lt;length&gt;</loc> values must be the same.</p>
 <p>The <loc href="#style-value-length">&lt;length&gt;</loc> value(s) used to express font size must be non-negative.</p>
+
+<p>Relative <loc href="#style-value-length">&lt;length&gt;</loc> values
```
skynavga: This and the following paragraph should match the presentation order in TTML2, i.e., go before the paragraph proscribing negative length values.

spec/ttml1.xml
```diff
@@ -4324,26 +4322,57 @@ glyphs do not have an EM square that is distinct from their font's EM square.</p
...
+
+
+<p>When a single relative <loc href="#style-value-length">&lt;length&gt;</loc> value is specified, then
+this <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the height component of the referenced value;
+when two relative
+<loc href="#style-value-length">&lt;length&gt;</loc> values are specified,
+the first <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the width component of the referenced value and
+the second <loc href="#style-value-length">&lt;length&gt;</loc> is resolved in terms of the height component of the referenced value.</p>
+
+<p>If the relative unit is cell (<code>c</code>), then the referenced value is the
```
skynavga, line 4381: The material added from here to line 4361 should be identical to the TTML2 text, but it is rephrased for no apparent reason.

skynavga (changes requested):
Without completing my detailed review, I have already provided new comments that clearly indicate the unnecessary departure from these changes and the TTML2 text that applies here as well. You need to start a new PR basing it exactly on the TTML2 changes and then propose (and argue for) specific changes that we should adopt both here and in TTML2.

spec/ttml1.xml
palemieux: The note placement is intended to match TTML1 2ED

spec/ttml1.xml
palemieux: Moved to match TTML1 placement

spec/ttml1.xml
palemieux: Modified to match TTML1

spec/ttml1.xml
palemieux: First sentence is already in `tts:fontSize` definition table, and the two following sentences are in fact in the prose.

spec/ttml1.xml
nigelmegitt: This text is in fact present in TTML2. The difference is that TTML2 introduces new text about resolving the c unit horizontally and vertically. I don't think that text is needed here at this time - if it is then please raise a separate issue. We're trying to close this down.

nigelmegitt (approved):
Approving the new changes - as agreed in today's WG meeting we will merge this since all the substantive comments have been addressed.  Any further issues should be raised separately after the merge.

palemieux:
@skynavga Please review per https://github.com/w3c/ttml1/pull/301#pullrequestreview-91605445 and revised PR

nigelmegitt:
@skynavga ping - last chance to re-review this please. Otherwise I'll have to dismiss your review so we can merge and make progress towards CR for TTML1 Third Ed. It's been 26 days since the WG agreed that all the substantive comments have been addressed - the only reason I'm not going ahead and doing this straight away is because I know you've been working on TTML2 issues recently.

skynavga:
@nigelmegitt give me until tomorrow

spec/ttml1.xml
skynavga: I would prefer that the order of the two notes "Anamorphic ..." and "A glyph's EM square ..." be changed so the latter appears first. This corresponds with the material in the preceding paragraph: EM square is used before referring to anamorphic (two values).

spec/ttml1.xml
```diff
@@ -4333,52 +4333,82 @@ as a style property only to those element types indicated in the following table
...
-the <emph>Computed Cell Size</emph>.
-When two &lt;length&gt; values are expressed using cell units, then the first refers to the width of
-the <emph>Computed Cell Size</emph>, and the second refers to the height of the
-<emph>Computed Cell Size</emph>.</p>
+the computed font size and the supported font size is minimized. If there are multiple closest supported values equally distant from		
+the computed value, then the value most distant from 0 (single length specification) or [0,0] (two length specifications) is 
+used, i.e., the largest font size, is used.</p>		
 </note>
+
+<p>If a computed value of the property associated with this attribute is not supported,
```
skynavga: This paragraph needs to go before the preceding note "In this context, the closest supported value ...", which purpose is to elaborate this paragraph (that now follows the note).

skynavga (changes requested):
I can approve once my two comments on ordering of notes is addressed.

palemieux:
@skynavga Implemented your suggestions. Please review.

skynavga (approved):
Thanks.

PR #302, opened 2017-12-20
Clarify definition of epoch and epoch-related offsets

### LINKED ISSUE #266: State that no epoch or epochOffset is defined
Issue split out from https://github.com/w3c/ttml1/issues/224#issuecomment-268787562

Somewhere in the timing attribute vocabulary section (possibly §10.2.4 timeContainer), add a statement along the lines of:

_The document processing context defines the applicable epoch and any epoch-related offsets to be used when establishing the synthetic document syncbase._

palemieux:
Close #266


nigelmegitt:
Built version visible at https://rawgit.com/w3c/ttml1/issue-266-epoch-definition-build/index.html

nigelmegitt (approved):
Thanks, looks good to me.

spec/ttml1.xml
```diff
@@ -13275,12 +13275,13 @@ the rate of playback (or interpretation) of time may artificially dilated or nar
...
 variable play rates, appropriate adjustments may need to be made to the resulting computations.</p>
 </note>
+<p>The <emph>Document Processing Context</emph> defines the applicable epoch and any epoch-related offsets to be used when establishing the <emph>Synthetic Document Syncbase</emph>.</p>
 <div2 id="time-expression-semantics-clock">
 <head>Clock Time Base</head>
 <p>When operating with the <code>clock</code> time base, the following semantics should be applied for interpreting time expressions,
 as defined by <loc href="#timing-value-timeExpression">&lt;timeExpression&gt;</loc>, and their relationship to media time and local real time.</p>
 <p>The clock time base <phrase role="strong"><code>C</code></phrase> is related to local real time <phrase role="strong"><code>R</code></phrase>
-expressed in an arbitrary (implementation defined) epoch <phrase role="strong"><code>E</code></phrase>
+expressed in the applicable epoch <phrase role="strong"><code>E</code></phrase>
```
cconcolato: Reading this sentence alone, one wonders what "the applicable epoch" is. Why not say "expressed in the epoch defined by the Document Processing Context".

nigelmegitt: That'd work, but the applicable epoch is now defined in the appendix text before the sub-sections begin, so I think it works as is.

skynavga: I would like to see the parenthetical "(implementation defined)" remain in some form.

skynavga (changes requested):

spec/ttml1.xml
palemieux: @skynavga Would the text suggested by @cconcolato address your comment since _Document Processing Context_ is implementation-defined?

skynavga: @palemieux yes, that would work

spec/ttml1.xml
```diff
@@ -13406,7 +13405,7 @@ during playback, presentation timing is not affected.
 <p>When operating with the <code>media</code> time base, the following semantics should be applied for interpreting time expressions,
 as defined by <loc href="#timing-value-timeExpression">&lt;timeExpression&gt;</loc>, and their relationship to media time and local real time.</p>
 <p>The media time base <phrase role="strong"><code>M</code></phrase> is related to local real time <phrase role="strong"><code>R</code></phrase>
-expressed in the applicable epoch <phrase role="strong"><code>E</code></phrase> as follows:</p>
+expressed in the epoch <phrase role="strong"><code>E</code></phrase> (defined by the <emph>Document Processing Context</emph>) as follows:</p>
```
skynavga: leave "applicable"; remove parenthesis; change to "as defined by ..."

skynavga (changes requested):

spec/ttml1.xml
palemieux: "as defined by" will be followed by "as follows", which I think reads funny.

skynavga: I thought you were going to use @cconcolato's text, which would be:

```
expressed in the epoch <phrase role="strong"><code>E</code></phrase> defined by the Document Processing Context as follows:
```

which is fine with me.

nigelmegitt: I think you're both wrong but also right! The point is not to give the impression that the "as follows" explains how the document processing context defines the epoch. How about:

_expressed in the applicable epoch E, where E is defined by the Document Processing Context, as follows:_

note the use of a comma after Document Processing Context to separate the sub-clause.

palemieux: I added the parentheses to emphasize that "as follows" qualifies epoch E and not _Document Processing Context_

nigelmegitt: I think proper parentheses are clearer here than commas. I'm still worried that the comma separated sub-clause could be misread.

skynavga:
I have reviewed this issue and PR and conclude that no change is needed whatsoever. The original text qualifies _epoch_ with "(implementation defined)", which means exactly what is intended. All of these efforts to point out that "(implementation defined)" means the same as "(document processing context defined)" serves no useful purpose and merely adds clutter to the document.

palemieux:
@skynavga Evidently other members feel differently. The question is whether you plan on formally objecting to the PR as currently presented. Recall that the text is purely informative.

nigelmegitt:
I'm happy with the changes made after my review approval.

andreastai (approved):
I think the latest change fulfills it's purpose and address the request of the commenter. Therefore we should move on and merge the commit.

skynavga (approved):
In the interest of moving forward, I am approving this change. However, I oppose this change and will not entertain it for TTML2. This change fails to take note that "implementation defined|dependent" is a term in use elsewhere in the document, and adds nothing that is not already stated.

PR #306, opened 2017-12-26
Make smooth transitions a possibility rather than a recommendation

### LINKED ISSUE #304: Smooth transitions between ISDs should not be preferred
See also w3c/ttml2#485

The note beneath the `[associate region]` algorithm states:

> Where an implementation is able to detect significant similarity between two adjacent synchronic document instances, DOCinterN DOCinterN−1, then it is preferred that the implementation make the transition between presenting the two instances as smooth as possible ...

The use of the term "preferred" is territory-specific and not culturally accepted globally, so it should either be removed or scoped to the places where it is true. For example in the US smooth scrolling is preferred, whereas in the UK jump scrolling is preferred, and those preferences are strong active preferences.

I propose to change the wording to: "..., then the implementation may apply processing to make the transition between presenting the two instances as smooth as possible..." to remove the preference.

I suspect @mikedo could be particularly interested in this, based on my recollection of previous conversations.

palemieux:
Closes #304


palemieux:
Spec at https://rawgit.com/w3c/ttml1/issue-304-smooth-transition-not-preferred-build/index.html

skynavga (changes requested):
I oppose this change, since the preference was explicitly part of a WG resolution in the past to resolve a CR comment. What we can do, however, is add an additional note clarifying that this preference applies for ITU Region 2, and that other regions may have different preferences.

nigelmegitt (approved):
Looks good to me.

nigelmegitt:
@skynavga Pointer to the WG resolution/decision here please? I do not think we need to mention ITU regions - it is up to the implementer to understand which regional preferences apply and therefore whether or not to implement this behaviour. Appending to the version of the text in this pull request something like "This behaviour is preferred in some regions." would be adequate.

palemieux:
>  is up to the implementer to understand which regional preferences apply and therefore whether or not to implement this behaviour. 

Moreover, smooth transitions are not applicable to all transitions and all circumstances.

skynavga (approved):

PR #318, opened 2018-01-04
Clarify anonymous span construction and implicit duration semantics

### LINKED ISSUE #193: Inconsistent implicit duration of singleton span in sequential container.
The implicit duration of an anonymous span child of a sequential container is explicitly defined to be zero (section 10.4); however, the implicit duration of an explicit singleton (i.e., non-nested) span in the same context 

Here is an example, where I ignore whitespace to aid in viewing. Given a fragment:

```xml
<p timeContainer="seq">
  Foo
  <span>Bar</span>
  Baz
</p>
```

Generating anonymous spans to wrap text nodes:

```xml
<p timeContainer="seq">
  <span xml:id="anon1">Foo</span>
  <span>
    <span xml:id="anon2">Bar</span>
  </span>
  <span xml:id="anon3">Baz</span>
</p>
```

Applying timing semantics as presently defined to resolve implicit duration:

```xml
<p timeContainer="seq" durImplicit="indefinite">
  <span xml:id="anon1" durImplicit="0">Foo</span>
  <span durImplicit="indefinite">
    <span xml:id="anon2" durImplicit="indefinite">Bar</span>
  </span>
  <span xml:id="anon3" durImplicit="0">Baz</span>
</p>
```

Here, anon2 is assigned an indefinite implicit duration because its parent is an explicit span which defaults to 'par' time containment semantics. Since SMIL requires a seq container that contains an indefinite duration child to also be indefinite, then that indefinite propagates up the node tree.

This seems counterintuitive. I propose to resolve by a slight change in 10.4 from the current text:

"if the anonymous span's parent time container is a parallel time container, then the implicit duration is equivalent to the indefinite duration value as defined by [SMIL 2.1]; if the anonymous span's parent time container is a sequential time container, then the implicit duration is equivalent to zero"

to read:

"if the anonymous span's nearest explicit time container ancestor is a sequential time container, then the implicit duration is equivalent to zero; otherwise, it is equivalent to the indefinite duration value as defined by [SMIL 2.1]"

and add the following Note

"Note: An element is an explicit time container if it specifies a timeContainer attribute."

Reworking the above example with the new text, we get:

```xml
<p timeContainer="seq" durImplicit="0">
  <span xml:id="anon1" durImplicit="0">Foo</span>
  <span durImplicit="0">
    <span xml:id="anon2" durImplicit="0">Bar</span>
  </span>
  <span xml:id="anon3" durImplicit="0">Baz</span>
</p>
```

The resolution of this issue also needs applied to TTML2.

(raised by Glenn Adams on 2014-08-26)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/338

### LINKED ISSUE #310: Implicit durations of <set> and <br> not specified
Section 10.4 does not specify the implicit duration of `br` and `set` elements.

I recommend using the same implicit duration as anonymous spans. Such an approach does not yield different results across the IMSC1 test vectors.

palemieux:
Closes #193 
Closes #310

### DIFF
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -8079,9 +8079,8 @@ is equivalent to zero.</p>
 </item>
 <item>
 <p>The implicit duration of a <el>body</el> element, <el>div</el> element, <el>p</el> element,
-or <el>span</el> element (but not of a <el>span</el> element whose children 
-are exclusively text nodes), is determined in accordance to (1) whether
-the element is a parallel or sequential time container, (2) the default
+or <el>span</el> element whose children are not exclusively text nodes, is determined in 
+accordance to (1) whether the element is a parallel or sequential time container, (2) the default
 <code>endsync</code> semantics defined above by <specref ref="timing-attribute-timeContainer"/>,
 and (3) the semantics of <bibref ref="smil21"/> as applied to these time
 containers.</p>
@@ -8110,10 +8109,19 @@ based timing or event based timing in <bibref ref="smil21"/>, where the specific
 <specref ref="time-expression-semantics-smpte"/> apply.</p>
 </item>
 </ulist>
-<p>For the purposes of performing presentation processing, the <phrase role="strong">[construct anonymous spans]</phrase> procedure is applied before resolving begin and end times of content elements.</p>
-<p>The following example illustrates the timing semantics of anonymous spans that are children of sequential time containers. The words <emph>Hello</emph> and <emph>Allo</emph> are both contained in anonymous spans that are children of a sequential time container (the outermost <code>p</code> element). As such, neither <emph>Hello</emph> nor <emph>Allo</emph> will be displayed since the implicit duration of the anonymous spans is 0. In contrast, the word <emph>Bonjour</emph> is contained within an anonymous span that is the child of a parallel time container (the innermost <code>p</code> element). As such, <emph>Bonjour</emph> will be displayed since the implicit duration of the anonymous span is indefinite, resulting in the implicit duration of its <code>span</code>, <code>p</code>, <code>div</code> and <code>body</code> ancestors to also be indefinite. </p>
+<p>For the purposes of performing presentation processing, the <phrase role="strong">
+[construct anonymous spans]</phrase> procedure is applied before resolving begin and end times of content elements.</p>
+<p>The following example illustrates the timing semantics of anonymous spans that are children of sequential and parallel time containers. 
+The words <emph>Hello</emph> and <emph>Allo</emph> are both contained in anonymous spans that are children of a sequential 
+time container (the <code>p</code> element). As such, neither <emph>Hello</emph> nor <emph>Allo</emph> will be displayed 
+since the implicit duration of the anonymous spans is 0. In contrast, the anonymous span that contains the word <emph>Guten</emph> 
+and the innermost <code>span</code> that contains the word <emph>Tag</emph> are children of an element with parallel time container 
+semantics (the outermost <code>span</code> element).
+As a result, both <emph>Guten</emph> and <emph>Tag</emph> will be displayed since the implicit duration of their respective parent anonymous span 
+and <code>span</code> element are indefinite, resulting in the implicit duration of 
+its <code>span</code>, <code>p</code>, <code>div</code> and <code>body</code> ancestors to also be indefinite.</p>
 <table id="timing-intervals-example-1" role="example">
-<caption>Example Fragment &ndash; Timing of Anonymous Spans in a Sequential Time Container</caption>
+<caption>Example Fragment &ndash; Timing of Anonymous Spans in a Sequential and Parallel Time Containers</caption>
 <tbody>
 <tr>
 <td>
@@ -8123,7 +8131,7 @@ based timing or event based timing in <bibref ref="smil21"/>, where the specific
   &lt;div&gt;
     &lt;p timeContainer=&quot;seq&quot;&gt;
       Hello
-      &lt;p&gt;Bonjour&lt;/p&gt;
+      &lt;span&gt;Guten &lt;span&gt;Tag&lt;/span&gt;&lt;/span&gt;
       Allo
     &lt;/p&gt;
   &lt;/div&gt;
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -3095,7 +3095,7 @@ followed by
 zero or more intermixed
 <code>span</code> elements,
 <code>br</code> elements,
-or text nodes interpreted as an anonymous spans.</p>
+or text nodes.</p>
 <p>Any metadata specified by children in the <code>Metadata.class</code>
 element group applies semantically to the <el>p</el> element and its descendants as a whole.</p>
 <p>Any animation elements specified by children in the <code>Animation.class</code>
@@ -3167,7 +3167,7 @@ followed by
 zero or more intermixed
 <el>span</el> elements,
 <el>br</el> elements,
-or text nodes interpreted as anonymous spans.</p>
+or text nodes.</p>
 <p>Any metadata specified by children in the <code>Metadata.class</code>
 element group applies semantically to the <el>span</el> element and its descendants as a whole.</p>
 <p>Any animation elements specified by children in the <code>Animation.class</code>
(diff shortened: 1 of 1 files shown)


cconcolato (commented):
Maybe changing "timing semantics" to "timing behavior"?

nigelmegitt:
Build available at https://rawgit.com/w3c/ttml1/issue-193-anon-span-implicit-duration-build/index.html

spec/ttml1.xml
```diff
@@ -8094,6 +8094,30 @@ based timing or event based timing in <bibref ref="smil21"/>, where the specific
 <specref ref="time-expression-semantics-smpte"/> apply.</p>
 </item>
 </ulist>
+<p>The following example illustrates the timing semantics of anonymous spans that are children of sequential time containers. The words <emph>Hello</emph> and <emph>Allo</emph> are both contained in anonymous spans that are children of a sequential time container (the <code>p</code> element). As such, neither <emph>Hello</emph> nor <emph>Allo</emph> will be displayed since the implicit duration of the anonymous spans is 0. In contrast, the word <emph>Bonjour</emph> is contained within an anonymous span that is the child of a parallel time container (the <code>span</code> element). As such, <emph>Bonjour</emph> will be displayed since the implicit duration of the anonymous span is indefinite, resulting in the implicit duration of its <code>span</code>, <code>p</code>, <code>div</code> and <code>body</code> ancestors to also be indefinite. </p>
```
nigelmegitt: I think the normal editorial practice is to put the explanation in a Note beneath the example.

nigelmegitt: I don't think it is true that the word "Bonjour" is contained in an anonymous span. The text in §7.1.5 p:

> If a sequence of children of a p element consists solely of character information items, then that sequence must be considered to be an anonymous span for the purpose of applying style properties that apply to span elements.

does not appear in 7.1.6 span so I don't think it applies.

Rather, the reason the word Bonjour appears indefinitely is because the explicit span's implicit duration is indefinite.

nigelmegitt (dismissed):
Requesting a formatting change but more importantly I don't agree with the given explanation for the behaviour (I do agree with the described behaviour though!).

spec/ttml1.xml
nigelmegitt: Please disregard the previous comment - I've gone and checked and read it again properly and realise that 7.1.6 does have text that provides this behaviour:

> The span element accepts as its children zero or more elements in the Metadata.class element group, followed by zero or more elements in the Animation.class element group, followed by zero or more intermixed span elements, br elements, or **text nodes interpreted as anonymous spans**.

(my emphasis added)

Sorry for the noise.

nigelmegitt (commented):
The formatting is a little inconsistent but I'm basically happy with the content here.

spec/ttml1.xml
cconcolato: @nigelmegitt In fact, I think you were right in your first thought. In [9.3.3](https://www.w3.org/TR/ttml1/#semantics-region-layout-step-2) it actually says:

> for each span element whose child is a single anonymous span, replace the anonymous span with its sequence of child text nodes;

palemieux: @cconcolato This clause applies when constructing ISDs, which happens after timing is resolved.

cconcolato: @palemieux I'm confused. The creation of 'anonymous spans' is only defined in [9.3.3](https://www.w3.org/TR/ttml1/#semantics-region-layout-step-2), therefore for me it does not make sense to talk about timing of an anonymous span, because as you say timing has been resolved already in 9.3.3.

cconcolato: @palemieux TTML2 has a note that covers that:
> Note: A compliant presentation processor is permitted to perform the previous step (anonymous span generation) and this step (style resolution) as post-processing steps while performing the [construct intermediate document] procedure, in which case each intermediate synchronic document produced by 11.3.1.3 Intermediate Synchronic Document Construction will contain anonymously generated spans and computed 

but I couldn't find this text in TTML1. Maybe this should be backported.

palemieux: I think we need to understand why that note was added to TTML2.

In fact, TTML2 changed the sentence _the span element accepts as its children zero or more elements in the Metadata.class element group, followed by zero or more elements in the Animation.class element group, followed by zero or more intermixed span elements, br elements, or text nodes interpreted as anonymous spans_ to remove the mention of anonymous spans.

This is a problem since TTML2 also states that _If no timeContainer attribute is specified on a span element, then it must be interpreted as having parallel time containment semantics._, which leaves the implicit duration of a span undefined if the span element does not contain anonymous spans.

nigelmegitt: It does seem that we should not have a discrepancy here between TTML1 and TTML2 if at all possible. I've added an agenda item for this, but perhaps it is something that can be covered in offline discussions before the F2F on Tuesday?

palemieux:
Notes on https://github.com/w3c/ttml1/pull/318/commits/e0935653c1dd897da9e3afc102102bd6745af61f :

- the implicit duration of `set` elements follows the semantics of anonymous spans
- anonymous span construction and pruning has been factored out
- the implicit duration of `span` elements that contain only text nodes follow the semantics of anonymous spans. This means that:
1. `<p><span>hello</span></p>` behaves the same as `<p>hello</p>` 
2. anonymous span construction and pruning can take place once for both timing resolution and style resolution
- mentions of text nodes being converted to anonymous spans are removed from `p` and `span` sections per TTML2
- the example in Section 10.4 was updated accordingly

spec/ttml1.xml
```diff
@@ -8055,16 +8068,19 @@ vocabulary defined by this specification, given the following
...
-duration value as defined by <bibref ref="smil21"/>; if the anonymous span's
+duration value as defined by <bibref ref="smil21"/>; if the span's
 parent time container is a sequential time container, then the implicit duration
 is equivalent to zero.</p>
 </item>
 <item>
-<p>The implicit duration of a <el>body</el>, <el>div</el>, <el>p</el>,
-or <el>span</el> element is determined in accordance to (1) whether
+<p>The implicit duration of a <el>body</el> element, <el>div</el> element, <el>p</el> element,
+or <el>span</el> element (but not of a <el>span</el> element whose children
```
nigelmegitt: Rather than the parentheses it would be consistent with the previous bullet, and clearer, to say "or `span` element whose children are not exclusively text nodes".

spec/ttml1.xml
```diff
@@ -8106,7 +8123,7 @@ based timing or event based timing in <bibref ref="smil21"/>, where the specific
   &lt;div&gt;
     &lt;p timeContainer=&quot;seq&quot;&gt;
       Hello
-      &lt;span&gt;Bonjour&lt;/span&gt;
+      &lt;p&gt;Bonjour&lt;/p&gt;
```
nigelmegitt: This is not valid TTML - a `p` cannot contain a `p`.

spec/ttml1.xml
```diff
@@ -8094,7 +8110,8 @@ based timing or event based timing in <bibref ref="smil21"/>, where the specific
 <specref ref="time-expression-semantics-smpte"/> apply.</p>
 </item>
 </ulist>
-<p>The following example illustrates the timing semantics of anonymous spans that are children of sequential time containers. The words <emph>Hello</emph> and <emph>Allo</emph> are both contained in anonymous spans that are children of a sequential time container (the <code>p</code> element). As such, neither <emph>Hello</emph> nor <emph>Allo</emph> will be displayed since the implicit duration of the anonymous spans is 0. In contrast, the word <emph>Bonjour</emph> is contained within an anonymous span that is the child of a parallel time container (the <code>span</code> element). As such, <emph>Bonjour</emph> will be displayed since the implicit duration of the anonymous span is indefinite, resulting in the implicit duration of its <code>span</code>, <code>p</code>, <code>div</code> and <code>body</code> ancestors to also be indefinite. </p>
+<p>For the purposes of performing presentation processing, the <phrase role="strong">[construct anonymous spans]</phrase> procedure is applied before resolving begin and end times of content elements.</p>
```
nigelmegitt: Is _all_ of [construct anonymous spans] performed before resolving begin and end times? Consider:

```
<p>
Wibble
<span begin="2s" end="3s">Wobble</span>
Wibble
</p>
```

Before 2s and after 3s, if you resolve the timing before step 2 of [construct anonymous spans] there are two contiguous anonymous spans that get replaced by a single anonymous span in step 2. Between 2s and 3s the anonymous spans are not contiguous.

If you resolve the timing after completing [construct anonymous spans] then the anonymous spans are never contiguous.

The next question is: does it ever make any difference? I think it makes no difference, but then I wonder what the point of step 2 is at all, and suspect I've missed something... Any further thoughts or ideas @palemieux @skynavga ?

nigelmegitt (changes requested):
One editorial point, one technical issue in the example that certainly needs to be fixed, and one possibly hypothetical or irrelevant question that I'd like to be more certain about before moving forwards with this change.

spec/ttml1.xml
palemieux: @nigelmegitt Thanks for the catch!

spec/ttml1.xml
palemieux: > but then I wonder what the point of step 2 is at all, and suspect I've missed something

FWIW, I think of step (2) as a clean-up pass that create a single anonymous span from the multiple text nodes that an XML element can have.

nigelmegitt:
Thanks @palemieux those changes fix the main issues. I'd like to cover https://github.com/w3c/ttml1/pull/318#discussion_r162199073 on the call later today, hopefully briefly, before I approve this pull.

nigelmegitt (approved):
See also the point discussed today ([minutes](https://www.w3.org/2018/01/18-tt-minutes.html#item13)); following that discussion, I'm approving this.

PR #321, opened 2018-01-10
Clarify validation of undefined items in TT namespaces

### LINKED ISSUE #251: Use of attributes in TTML namespace but NOT defined in TTML1
Content models of TTML elements are defined as Element Information Items and represented in XML. One example is the content model of the `style` element:

```
<style
  style = IDREFS
  xml:id = ID
  xml:lang = string
  xml:space = (default|preserve)
  {any attribute in TT Style namespace}
  {any attribute not in default or any TT namespace}>
 Content: EMPTY
</style>
```
The content models give a "general allowance" to use attributes of a specific namespace. In the above example, any attribute that is in the TT Style namespace ("http://www.w3.org/ns/ttml#styling") can be used. 

My question is, if this is restricted to just the attributes of "http://www.w3.org/ns/ttml#styling" that are defined in TTML1 (see [Section 5.3.1 of TTML1, Table 5](https://www.w3.org/TR/ttaf1-dfxp/#core-vocabulary-overview)), if also attributes are allowed that have been defined for the TTML Style namespace in a later publication of W3C (e.g. TTML2) or if from the element type defintion it doesn't matter where attribute is defined, it just needs to be in the TTML Style namespace.

palemieux:
Closes #251

### DIFF
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -1098,7 +1098,10 @@ constrained by the author or authoring tool to satisfy a more
 restrictive definition of validity.</p>
 </note>
 <note>
-<p>As illustrated in the following example, an <emph>Abstract Document Instance</emph> can be a <emph>Valid Abstract Document Instance</emph> even if it includes elements and attributes whose namespace names are listed in <specref ref="namespace-vocab-table"/> but are not defined in this version of the specification. Specifically, the element <code>foo</code> and the attribute <att>tts:foo</att> are pruned by above steps (1) and (3), respectively, because they are not members of the associated <emph>Abstract Document Type</emph>, even though their namespace names are listed in <specref ref="namespace-vocab-table"/>.
+<p>As illustrated in the following example, an <emph>Abstract Document Instance</emph> can be a <emph>Valid Abstract Document Instance</emph> even if it includes elements and attributes whose namespace names are listed in <specref ref="namespace-vocab-table"/> but whose local names 
+are not part of the vocabulary defined by this version of the specification. Specifically, the element <code>foo</code> and the attribute <att>tts:foo</att> 
+are pruned by above steps (1) and (3), respectively, because they are not members of the associated <emph>Abstract Document Type</emph>, even though their 
+namespace names are listed in <specref ref="namespace-vocab-table"/>.
 </p>
 <table id="document-types-example" role="example">
 <caption>Example Fragment &ndash; Pruning of elements and attributes that are not members of the associated <emph>Abstract Document Type</emph></caption>
@@ -1162,7 +1165,7 @@ exceptional reporting of this false negative does not impact the formal assessme
 of <emph>Document Instance</emph> validity.</p>
 </note>
 <note role="clarification">
-<p>The schemas referenced by this specification are designed for use after the pruning steps (1)-(3) specified by <specref ref="doctypes"/> have been applied.</p>
+<p>The schemas referenced by this specification are intended for use after the pruning steps (1)-(3) specified by <specref ref="doctypes"/> have been applied.</p>
 </note>
 </div2>
 </div1>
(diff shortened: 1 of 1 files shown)


nigelmegitt:
Built version available at https://rawgit.com/w3c/ttml1/issue-251-ttml-ns-attribute-extension-build/index.html

Diff at https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fttml1%2Findex.html&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fttml1%2Fissue-251-ttml-ns-attribute-extension-build%2Findex.html

spec/ttml1.xml
```diff
@@ -1097,6 +1097,35 @@ conformant <emph>Document Instance</emph>, a given <emph>Document Instance</emph
 constrained by the author or authoring tool to satisfy a more
 restrictive definition of validity.</p>
 </note>
+<note>
+<p>As illustrated in the following example, an <emph>Abstract Document Instance</emph> can be a <emph>Valid Abstract Document Instance</emph> even if it includes elements and attributes whose namespace names are listed in <specref ref="namespace-vocab-table"/> but are not defined in this version of the specification. Specifically, the element <code>foo</code> and the attribute <att>tts:foo</att> are pruned by above steps (1) and (3), respectively, because they are not members of the associated <emph>Abstract Document Type</emph>, even though their namespace names are listed in <specref ref="namespace-vocab-table"/>.
```
nigelmegitt: How can an element or attribute have a namespace name that is both listed in the table and not defined in the specification? Is that supposed to say that its namespace is listed in the table but its name is not defined within that namespace by this specification?

spec/ttml1.xml
```diff
@@ -1097,6 +1097,35 @@ conformant <emph>Document Instance</emph>, a given <emph>Document Instance</emph
...
+          Bar
+      &lt;/p&gt;
+    &lt;/div&gt;
+  &lt;/body&gt;
+&lt;/tt&gt;
+</eg>
+</td>
+</tr>
+</tbody>
+</table>
```
nigelmegitt, line 1130: It seems to be more common in the spec to explain the example in following text rather than preceding text. Suggest moving the explanation to here.

spec/ttml1.xml
```diff
@@ -1132,6 +1161,9 @@ which all foreign namespace elements and attributes have been removed. Therefore
 exceptional reporting of this false negative does not impact the formal assessment
 of <emph>Document Instance</emph> validity.</p>
 </note>
+<note role="clarification">
+<p>The schemas referenced by this specification are designed for use after the pruning steps (1)-(3) specified by <specref ref="doctypes"/> have been applied.</p>
```
nigelmegitt: I think I'd prefer "intended" rather than "designed" here but that's a minor point.

nigelmegitt (changes requested):
The explanatory paragraph needs some editing.

spec/ttml1.xml
andreastai: Edited comment: Removed comments on the use of the term "namespace name" after checking the namespace spec.

andreastai (commented):
From my side this PR can be merged once the ambiguous use of "namespace name" is removed.

I will also not object to the PR as the validation strategy has been in TTML 1 for a long time.

spec/ttml1.xml
palemieux: > Is that supposed to say that its namespace is listed in the table but its name is not defined within that namespace by this specification?

Yes.

andreastai (approved):

nigelmegitt (approved):
I'd still prefer the explanation to come beneath the fragment example but it's not a serious enough problem to block approval.

skynavga:
This change may need modification to match what gets written into TTML2 on this subject.

PR #328, opened 2018-01-16
Improve overflow example

### LINKED ISSUE #320: Add better overflow example
See https://github.com/w3c/ttml1/issues/239#issuecomment-350071676

palemieux:
Closes #320


spec/examples/ex4-overflow.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<tt xmlns="http://www.w3.org/ns/ttml" xmlns:tts="http://www.w3.org/ns/ttml#styling" xml:lang="en" tts:extent="320px 240px">
+  <head>
+    <styling>
+      <style xml:id="s1" tts:backgroundColor="black" tts:wrapOption="noWrap" tts:padding="6px"/>
+      <style xml:id="s2" tts:color="red" tts:backgroundColor="purple" tts:fontFamily="proportionalSansSerif" tts:fontSize="10px"/>
```
nigelmegitt: I'm finding the colour contrast between the purple and the red really horrible (and the luminance contrast is very low too) - please could you change one of them, e.g. change the purple to grey?

spec/examples/ex4-overflow.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0"?>
+<tt xmlns="http://www.w3.org/ns/ttml" xmlns:tts="http://www.w3.org/ns/ttml#styling" xml:lang="en" tts:extent="320px 240px">
+  <head>
+    <styling>
+      <style xml:id="s1" tts:backgroundColor="black" tts:wrapOption="noWrap" tts:padding="6px"/>
```
nigelmegitt: Why put `tts:wrapOption` in the region style? It seems inconsistent to do this but put `tts:fontFamily` and `tts:fontSize` in the p style - those three should probably all go on the same style, I don't mind which one.

nigelmegitt (changes requested):
Please could you add a validating step to `build.xml` to check the new example XML file?

Also requesting a colour change and less importantly a change to make the application of styles consistent.

spec/examples/ex4-overflow.xml
palemieux: Grey does not provide sufficient contrast with the background. Please pick another color.

nigelmegitt: White?

spec/build.xml
```diff
@@ -179,6 +179,22 @@
...
+    <schemavalidate fullchecking="true" warn="true">
+      <schema namespace="http://www.w3.org/ns/ttml" file="${xsd.schema}"/>
+      <fileset dir="${examples.dir}">
+        <include name="ex4-overflow.xml"/>
+      </fileset>
+    </schemavalidate>
+  </target>
 
   <target name="validate-examples"
 	    depends="validate-example-1, validate-example-2, validate-example-3"/>
```
nigelmegitt: Need to add `validate-example-4` to the dependencies here.

nigelmegitt: (also, separately, I see that the `build` target does not call `validate` first, which it should)

nigelmegitt (changes requested):
Sorry, one more fix to make.

nigelmegitt (approved):
Thank you!

PR #330, opened 2018-01-16
Discourage animation class children on <br>

### LINKED ISSUE #325: Discourage use of Animation class children elements of <br>
`br` elements allow Animation class children but does not support timing attribute vocabulary

palemieux:
Closes #325


nigelmegitt (approved):
This works for me.

PR #331, opened 2018-01-16
Clarify tts:direction on <p>

### LINKED ISSUE #324: Backport text from TTML2 about the use of tts:direction at paragraph level
That sets the paragraph embedding level.
[from discussion at Jan. 2018 F2F, see #247]

palemieux:
Closes #324


nigelmegitt (commented):
I'm not able to review this because I feel my knowledge of Unicode bidi embedding is not good enough yet.

palemieux:
@nigelmegitt It is intended to match TTML2 exactly.

nigelmegitt (approved):
This change matches the equivalent text in TTML2.

palemieux:
Looks like this will need to be updated to match https://github.com/w3c/ttml2/issues/264

palemieux:
@nigelmegitt Updated per https://github.com/w3c/ttml2/pull/562. Please review.

spec/ttml1.xml
```diff
@@ -3755,9 +3755,8 @@ represented by this attribute are based upon that defined by <bibref ref="xsl11"
 </div3>
 <div3 id="style-attribute-direction">
 <head>tts:direction</head>
-<p>The <att>tts:direction</att> attribute is used to specify a style property that
-defines the directionality of an embedding or override according to
-the Unicode bidirectional algorithm.</p>
+<p>The <att>tts:direction</att> attribute is used to specify a style property that, depending on the context of use,
+determines (1) the bidirectional paragraph level, or (2) the directionality of a bidirectional embedding or override.</p>
```
nigelmegitt, line 3759: There's an extra clause in https://github.com/w3c/ttml2/pull/562/files#diff-5ac39ab39265717d36dfd789d254ce1aR8859 that should probably be brought in here.

spec/ttml1.xml
```diff
@@ -11610,6 +11614,12 @@ Agent Accessibility Guidelines 1.0</titleref>, W3C Recommendation, 17
 December 2002. (See
 <xspecref href="http://www.w3.org/TR/2002/REC-UAAG10-20021217/">http://www.w3.org/TR/2002/REC-UAAG10-20021217/</xspecref>.)
 </bibl>
+<bibl id="uax9" key="UAX9">M. Davis, A. Lanin, and A. Glass, <titleref
+href="http://www.unicode.org/reports/tr9/tr9-35.html">Unicode Bidirectional
```
nigelmegitt: This should be updated to tr9-37, and the date changed also.

nigelmegitt (changes requested):
Needs a couple of further tweaks to match w3c/ttml2#562.

spec/ttml1.xml
palemieux: @nigelmegitt I was not happy with the English "about which see". It seemed easiest to remove since UAX9 is already referenced below.

nigelmegitt: @palemieux I don't see anything wrong with "about which see", but I see what you mean about UAX9 being mentioned later, and I agree the clause is not essential here.

nigelmegitt (approved):

PR #333, opened 2018-01-19
Validate everything when building

### LINKED ISSUE #332: Make validate a dependency on the build target
The build target does not depend on `validate` however some of the files validated (such as the examples) find their way into the html spec output. We should make sure that `build` depends on `validate` to catch any validation errors introduced by pull requests.

nigelmegitt:
Closes #332


cconcolato (approved):
Same as TTML2 build

PR #319, opened 2018-01-05
Add Media timing section

nigelmegitt:
Closes #300 by applying the same changes as in w3c/ttml2#143 with the exception that terminology references are in `<emph>` elements rather than navigable `<loc>` elements.

PR #334, opened 2018-01-22
Clarify relationship between media time expressions and media times

### LINKED ISSUE #300: Clarify timing concepts in relation to SMIL
See w3c/ttml2#135 and pull request https://github.com/w3c/ttml2/pull/143 that led to [TTML2 §12.4.1 Media Timing](https://w3c.github.io/ttml2/spec/ttml2.html#semantics-media-timing) 

Should we also apply the same edits to TTML1 Third Edition? 

I see no reason why not. The SMIL reference would need to be made consistent with the TTML1 SMIL 2.1 reference, unless we also make TTML1 Third Edition reference SMIL 3. The rest of the section is equally true and useful for TTML1 as for TTML2.

palemieux:
Closes #300 
Intended to replace https://github.com/w3c/ttml1/pull/319


spec/ttml1.xml
```diff
@@ -13667,6 +13676,87 @@ and <code>tickRate</code> is the computed value of the <att>ttp:tickRate</att> p
 <p>The above formalisms assumes that the <emph>Root Temporal Extent</emph> corresponds with the beginning of a related media object. If
 this assumption doesn't hold, then an additional offset that accounts for the difference may be introduced when computing media time <code>M</code>.</p>
 </note>
+</div3>
+
+<div3 id="semantics-media-timing">
```
nigelmegitt: The section above here titled "TTML Semantics – Time Expressions and Media Time Relationship" looks like it should belong within this `semantics-media-timing` div3, or possibly be in its own div3. The change here puts it within the **N.2.1 Relationship to Local Real Time** section which that looks strange since it is orthogonal to that. @palemieux would you mind putting it in its own div3?

palemieux:
Build at http://rawgit.com/w3c/ttml1/issue-0300-v2-media-timing-semantics-build/index.html

spec/ttml1.xml
palemieux: See proposed text.

nigelmegitt (approved):
Thanks, I think this works. The last note in N.2.2 suggests an offset, and the first statement in N.2.3 says there is no offset, so that could be slightly confusing, but I think it's correct in the end

palemieux:
Tried to close/open to restart Travis build.

PR #346, opened 2018-03-01
Reorder features in profile documents

### LINKED ISSUE #203: The ordering of features in Appendix D differs from that in Appendix F.
The task of looking up  the features listed in each of the profiles in Appendix F against their definitions in Appendix D is made more difficult than it needs to be because the feature designators are listed in a different order.

For example #clockMode appears before #clockMode-gps in Appendix D but after #clockMode-utc in Appendix F.

The request is to use the same ordering in both sections. My personal preference is to retain the order in Appendix D but it isn't particularly important which is used. Lexical ordering of the feature designators probably makes most sense.

(raised by Nigel Megitt on 2015-08-18)
From tracker issue http://www.w3.org/AudioVideo/TT/tracker/issues/403

nigelmegitt:
Closes #203 by making the profile documents ordered primarily by `@value` and then by lexical order of feature designator, which matches Appendix D.


skynavga (approved):
LGTM

PR #353, opened 2018-05-10
Move edition number to subtitle (#352)

### LINKED ISSUE #352: Consider moving (Third Edition) to a subtitle
Per https://github.com/tobie/specref/issues/463, I suggest considering moving  (Third Edition) to a subtitle

palemieux:
Closes #352

### DIFF
--- a/spec/entities.dtd
+++ b/spec/entities.dtd
@@ -1,14 +1,15 @@
 
 <!-- comment / uncomment the appropriate entity -->
-<!-- <!ENTITY % sub.entities SYSTEM "entitiesedcopy.dtd" > -->
+<!ENTITY % sub.entities SYSTEM "entitiesedcopy.dtd" >
 <!-- <!ENTITY % sub.entities SYSTEM "entitieswd.dtd" > -->
-<!ENTITY % sub.entities SYSTEM "entitiescr.dtd" >
+<!-- <!ENTITY % sub.entities SYSTEM "entitiescr.dtd" > -->
 <!-- <!ENTITY % sub.entities SYSTEM "entitiesper.dtd" > -->
 <!-- <!ENTITY % sub.entities SYSTEM "entitiesrec.dtd" > -->
 %sub.entities;
 
 <!-- Titles -->
-<!ENTITY title "Timed Text Markup Language 1 (TTML1) (Third Edition)" >
+<!ENTITY title "Timed Text Markup Language 1 (TTML1)" >
+<!ENTITY subtitle "Third Edition" >
 
 <!-- XML Namespaces used in the drafts -->
 <!ENTITY nsuri "http://www.w3.org/ns/ttml" >
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -6,6 +6,7 @@
 <!ENTITY status SYSTEM "status.xml">
 <!ENTITY document.status "W3C Editors' Draft">
 <!ENTITY title "&title;">
+<!ENTITY subtitle "&subtitle;">
 <!ENTITY prevloc "http://www.w3.org/TR/2013/REC-ttml1-20130924/">
 <!ENTITY latestedloc "https://w3c.github.io/ttml1/index.html">
 <!ENTITY versionOfTT "1">
@@ -51,6 +52,7 @@
 <spec w3c-doctype="&doctype;" role="&document.role;">
 <header>
 <title>&title;</title>
+<subtitle>&subtitle;</subtitle>
 <w3c-designation>&w3c-designation;</w3c-designation>
 <w3c-doctype>&document.status;</w3c-doctype>
 <pubdate>
--- a/spec/xmlspec-ttml1.xsl
+++ b/spec/xmlspec-ttml1.xsl
@@ -498,7 +498,7 @@ width: 100%;
           <xsl:with-param name="conditional" select="0"/>
           <xsl:with-param name="default.id" select="'subtitle'"/>
         </xsl:call-template>
-        <xsl:apply-templates select="subtitle"/>
+        <xsl:value-of select="subtitle"/>
       </h2>
     </xsl:if>
     <xsl:text>&#10;</xsl:text>


skynavga (changes requested):
If IMSC is changed to use version specific references, then this change is not needed. Therefore, we cannot approve unless the IMSC reference issue is resolved first.

palemieux:
@skynavga While this issue was triggered by an issue in IMSC 1.0.1, it is orthogonal. The proposal is to separate the title of the specification from its version/edition numbers at the HTML level -- both will still appear together when displayed (see https://rawgit.com/w3c/ttml1/issue-352-move-edition-num-to-subtitle-build/index.html) . As discussed during the call, this is intended to facilitate automation and is consistent with W3C practice.

skynavga:
@palemieux I understand, but if the only reason is to become compatible with SOME novel W3C practice, then that is not a sufficient reason. The entire motivation for this change goes away if IMSC is fixed to use a specific reference. Backward compatibility with titling is more important than a change that serves no purpose except doing the "latest thing". If you show me the W3C POLICY RULE that requires such a change, I will be satisfied to make it.

nigelmegitt:
> The entire motivation for this change goes away if IMSC is fixed to use a specific reference.

I disagree, the motivation is to make use of proposed changes to specref or to W3 backend systems so that _anyone_ using the generic "latest" reference to TTML1 can obtain a generic unversioned title whilst those using a dated reference can get the fully qualified title. (in other words I'm agreeing with https://github.com/w3c/ttml1/pull/353#issuecomment-388254819)

The question of which reference to use in IMSC is orthogonal and is open at w3c/imsc#381.

> Backward compatibility with titling is more important than a change that serves no purpose except doing the "latest thing".

Why, what actual benefit does backward compatibility confer? @skynavga can you point to any systems that _parse_ the title and that would be broken by this change, as opposed to those that just copy the title? The same information will remain available on the spec page for human eyes. I think we already have a commitment that W3 will find a way to fix it so that specref gets the correct title based on the URL, but that can only work if we make this change.

> If you show me the W3C POLICY RULE that requires such a change, I will be satisfied to make it.

@skynavga you're setting the bar too high here - I would not expect a policy statement from W3C with this level of specificity.

nigelmegitt:
Build available at https://rawgit.com/w3c/ttml1/issue-352-move-edition-num-to-subtitle-build/index.html

nigelmegitt:
A further point to make here is that the current title does not actually include the date: that information is extracted from a separate entity in the document. So there is a structural precedent for separating out the elements of the full title as would be referenced.

In other words, the full reference of the built version for an ED based on this pull request might be quoted as "Timed Text Markup Language 1 (TTML1), Third Edition, W3C Editors' Draft 10 May 2018". The title and the document maturity level plus the date are already extracted from separate entities, so why not push this one step further and extract the edition from a third entity?

skynavga:
My position is as follows:

1. Absent a requirement for a generic title from IMSC (assuming it will be modified to use a specific reference), the is NO PROBLEM to solve here.
2. There is no W3C STYLE requirement or authoring policy that drives this change.
3. Any other reference to a requirement of an imagined document using specref or some other imagined tool set is PURE SPECULATION.
4. Driving style changes in order to satisfy one particular tool implementation (specref) is BAD POLICY, particularly when it would be simple enough to post a PR to specref that makes in more flexible.

My conclusion is: this PR is proposing CHANGE FOR CHANGE'S SAKE. I therefore FORMALLY OBJECT.

nigelmegitt:
> 2. There is no W3C STYLE requirement or authoring policy that drives this change.
> 3. Any other reference to a requirement of an imagined document using specref or some other imagined tool set is PURE SPECULATION.

@plehegar may be able to add something here, but I don't believe either of these is a correct statement. @plehegar introduced the following to us at TPAC 2017:

There has been a significant initiative in W3C to be able to handle generic URLs for specifications as well as specific ones, and to be able to serve meaningful pages for URLs such as https://www.w3.org/TR/ttml/all and https://www.w3.org/TR/ttml as well as https://www.w3.org/TR/ttml1 where the document that is served for any one of those URLs might change over time to reflect updates and other developments.

To my mind that initiative provides the context for this pull request, with a consequence that we need a title that stays consistent across changes to the page served for a URL. In some ways this pull request does not go far enough. There's a hierarchy like:

```
Timed Text Markup Language (TTML) (top level "brand")
 |-- TTML1
 |  |-- First edition
 |  |-- Second edition
 |  |-- Third edition
 |
 |-- TTML2
    |-- (implicitly) First edition
```

and then within each of those there are maturity levels, published specification dates and even in-place edit publication dates.

If we are to support this initiative (and we are!) we need a pull request like this in order to allow humans and systems to navigate this tree.

The only link to IMSC is that IMSC may have inadvertently alerted us to the problem - it did not create the problem, and by the way, merging this pull request will not fix the issue in IMSC either.

skynavga:
@nigelmegitt I'm willing to find a compromise provided that the final text of the title of TTML1 3rd Edition is:
```
Timed Text Markup Language 1 (TTML1) (Third Edition)
```

nigelmegitt:
@skynavga thanks for the willingness to compromise. I don't understand how your proposal to put "Third Edition" in parenthesis solves the issue in a technical sense? Are you envisaging that systems will have to parse the title of the spec against some kind of regular expression to extract the relevant subsections? I can't see that working very well.

skynavga:
I want the title to remain syntactically identical with 2nd and 1st edition titles; I don't care what other systems or documents do to obtain some either more or less specific, e.g., add a date,  remove the edition, etc. I view that to be a problem outside the scope of the TTML specification itself. However, I would not object to adding hidden metadata, e.g., `<meta>` elements, in the published HTML to facilitate such an external process.

nigelmegitt:
> I want the title to remain syntactically identical with 2nd and 1st edition titles

why?

skynavga:
Consistency.

nigelmegitt:
> Consistency.

@skynavga earlier you objected to change for change's sake. One could also argue that you want consistency for consistency's sake. This is only presentational.

skynavga:
@nigelmegitt since you have studied mathematics, you know there is a big difference between maintaining continuity and creating discontinuity; further, it is well known that the former is always preferred over the latter unless there are overwhelming reasons to BREAK the status quo; IMO, neither the current reference problem with IMSC nor the hypothetical future reference problem is an overwhelming reason to BREAK the status quo

nigelmegitt:
@skynavga I think you're making a lot of noise about what for humans is a minor presentational difference and for machines is a significantly helpful change. This has been argued sufficiently, we don't need to try going round the loop again.

I'm putting up with this for the time being pending any greater certainty about the implementation of W3 system support for subtitle in `h2` - when we have confirmation that such a change is going ahead and will use the subtitle in `h2` as a data source I hope you'll withdraw your (capitalised or not) formal objection so that we can go ahead and merge this with consensus, otherwise as Chair I'll have to make a call on proceeding in the presence of the objection. 

@plehegar I'm going to assign this to you: please could you post to this issue any further details about planned backend support in W3 for (not) using subtitle in h2 as part of the title depending on whether a generic title or an edition-specific title is needed? Also if there is any particular urgency in making the change now?

PR #355, opened 2018-07-05
Update tts:lineHeight='normal' algorithm to match TTML2

### LINKED ISSUE #354: Computed value of tts:lineHeight="normal" should match TTML2 CR2
See https://www.w3.org/TR/ttml2/#style-attribute-lineHeight

palemieux:
Closes #354

### DIFF
--- a/spec/ttml1.xml
+++ b/spec/ttml1.xml
@@ -4716,15 +4716,43 @@ as a style property only to those element types indicated in the following table
 </tbody>
 </table>
 <p>If the value of this attribute is <code>normal</code>, then the
-computed value of the style property must be considered to be no less than the
-size of the largest font size that applies to the element and its
-descendant elements in the intermediate synchronic document as determined by
-<specref ref="semantics-region-layout-step-1"/>.</p>
+resolved computed value of this style property should be determined as follows:</p>
+<olist>
+<item><p>
+Let <emph>P</emph> be the <loc href="#content-vocabulary-p"><el>p</el></loc> element to which this style property applies.
+</p></item>
+<item><p>
+Let <emph>FF</emph> be the computed value of the <loc href="#style-attribute-fontFamily"><att>tts:fontFamily</att></loc>
+style property of <emph>P</emph>.
+</p></item>
+<item><p>
+Let <emph>FS</emph> be the computed value of the <loc href="#style-attribute-fontSize"><att>tts:fontSize</att></loc> style property of <emph>P</emph>.
+</p></item>
+<item><p>
+Let <emph>F0</emph> be the first font obtained when sequentially mapping each font family in <emph>FF</emph> to a set of available fonts,
+where this set of available fonts is constrained as needed to satisfy the computed values of the
+<loc href="#style-attribute-fontStyle"><att>tts:fontStyle</att></loc> and 
+<loc href="#style-attribute-fontWeight"><att>tts:fontWeight</att></loc>
+style properties of <emph>P</emph>.
+</p></item>
+<item><p>
+If <emph>F0</emph> is associated with font metrics that specify altitude <emph>A</emph>, descent <emph>D</emph>, and line gap <emph>G</emph>,
+then set <emph>LH</emph> to the sum of scaled(<emph>A</emph>), scaled(<emph>D</emph>), and scaled(<emph>G</emph>),
+where scaled(<emph>X</emph>) denotes font metric <emph>X</emph> scaled according to font size <emph>FS</emph>.
+</p></item>
+<item><p>Otherwise, <emph>LH</emph> is considered to be implementation dependent; however, in the absence of implementation specific requirements,
+<emph>LH</emph> is recommended to be set to 125% of <emph>FS</emph>.
+</p></item>
+<item><p>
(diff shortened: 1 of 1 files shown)


skynavga (approved):

spec/ttml1.xml
```diff
@@ -4716,15 +4716,43 @@ as a style property only to those element types indicated in the following table
...
+<emph>LH</emph> is recommended to be set to 125% of <emph>FS</emph>.
+</p></item>
+<item><p>
+Set the resolved computed value of this style property to <emph>LH</emph>.
+</p></item>
+</olist>
 <note role="elaboration">
-<p>It is recommended that presentation processors use a computed line height that is
-125% of the size of the largest applicable font size when the value <code>normal</code>
-is specified. The choice of 125% allows use of a font size of <code>0.8c</code> in order
```
nigelmegitt, line 4726: We've lost the note text that points out that 125% and 0.8c go well with each other, both here and from TTML2. It'd be useful to add it back under the recommendation to use 125% above.

spec/ttml1.xml
```diff
@@ -4716,15 +4716,43 @@ as a style property only to those element types indicated in the following table
 </tbody>
 </table>
 <p>If the value of this attribute is <code>normal</code>, then the
-computed value of the style property must be considered to be no less than the
-size of the largest font size that applies to the element and its
-descendant elements in the intermediate synchronic document as determined by
-<specref ref="semantics-region-layout-step-1"/>.</p>
+resolved computed value of this style property should be determined as follows:</p>
```
nigelmegitt, line 4719: Now that the text mirrors the TTML2 language use of "resolved computed value" I think we also need the note from TTML2 about used value and inherited value, and we can then drop the special inheritance semantics subsection.

nigelmegitt (changes requested):

spec/ttml1.xml
skynavga: I see no reason to vary from the TTML2 text, so I oppose restoring that note.

palemieux: I do not see the note in TTML2 re: `0.8c`. I also think it is misleading since there is simply no guarantee that `125%` will be used. Also, why is `0.8c` relevant, I forget.

spec/ttml1.xml
palemieux: TTML2 was intended to include the `special inheritance semantics` section, but it was missed before CR2, so AFAIK the TTML1 text is preferred.

I can revert "resolved computed value" to "computed value" to match CR1 if you prefer.

skynavga: That would be factually wrong, i.e., to use _computed value_, because the computed value is the inherited value, but the result of this computation is not the inherited value.

spec/ttml1.xml
nigelmegitt: I would like to add it back into TTML2 also.

0.8c is relevant because (legacy) grid based systems don't make a distinction between font size and line height, so there is often a need to "fit text within a row", i.e. the line height needs to end up being 1c and a font size needs to be chosen that matches that. Hence the likely application of fontSize = 0.8c lineHeight = 125% which matches arithmetically without the need for rounding. Contrast with lineHeight = 120% which would require a fontSize of 0.833333...

spec/ttml1.xml
nigelmegitt: My reading of the TTML2 language is that it bypasses the need for a special section on inheritance and that as it stands in this pull request the special inheritance semantics don't apply.

palemieux: @nigelmegitt The following note was added to TTML2 instead of the special inheritance section. I suggest we keep the normative special inheritance section.

_When the specified value of this attribute is normal, the computed value is the specified value; therefore, when applying inheritance semantics, the value normal is the inherited value. However, when this computed value is to be used to perform paragraph layout, it is further processed to obtain a resolved computed value (also called a used value) as described above._

spec/ttml1.xml
palemieux: @nigelmegitt Ok. However, as discussed before, the 125% is only a suggestion, and setting the fontsize to 0.8c is no guarantee that the actual line height will be exactly 1c.

andreastai:
My review of the algorithm comes late. So, my comments are non-blocking to not interfere with the publication process.

I have some concerns. The recommended algorithm assumes that a presentation processor has access to the font properties of a font file when it calculates the resolved computed value. I doubt that this is the case.

This is also not an algorithm defined in CSS. To be more compliant with CSS one option would be to take the (very vague) text from CSS 2.1.

If a change is not possible, some terms in the algorithm may need to be clarified. What is the reference for the terms `altitude` and `descent`? To be clear some terms from a font format as open type should be used instead. This can be then used as reference for equivalent font formats.

nigelmegitt (approved):
Based on the TTWG meeting discussion of 2018-07-12, I'm approving this now. Any further changes made in TTML2 will be editorial in nature, and need to be back-ported to TTML1 3rd Ed also, but do not need to hold this pull request up, since editorial changes can be made later.

PR #356, opened 2018-07-05
Clarify the default value of blur radius for tts:textOutline

### LINKED ISSUE #351: Clarify default value of blur radius
The default value of blur radius is not specified. It should be clarified to be zero.

palemieux:
Closes #351


skynavga (approved):

nigelmegitt (approved):
Matches TTML2 CR2.

PR #375, opened 2020-05-27
Remove bad character

MarcAntoine-Arnaud:


cconcolato:
I don't understand where the bad character is. Can you explain?

MarcAntoine-Arnaud:
3 bytes at the start. Invisible on text editors.
You can see them with an hex dump ...

MarcAntoine-Arnaud:
You can see here, 3 first bytes with values: `0xef 0xbb 0xbf` precede the XML content

Screenshot 2020-05-28 at 10 50 23

himorin:
ef bb bf is BOM for utf-8 text. I don't think we need this change.

MarcAntoine-Arnaud:
Can make sense, but it's the only XSD with that information...

MarcAntoine-Arnaud:
I have read more about BOM. And I have handled it in my tool.
It's the first time I found that, very interesting but not really common.

PR #378, opened 2021-03-22
updated DTD/XSL

himorin:
to match with recent /TR/ spec template update, contents as https://github.com/w3c/ttml2/pull/1224/files

for https://github.com/w3c/ttwg/issues/181

### DIFF
--- a/spec/xmlspec-ttml1.xsl
+++ b/spec/xmlspec-ttml1.xsl
@@ -478,12 +478,6 @@ width: 100%;
     </xsl:if>
     <xsl:text>&#10;</xsl:text>
     <h1>
-      <xsl:call-template name="anchor">
-        <xsl:with-param name="node" select="title[1]"/>
-        <xsl:with-param name="conditional" select="0"/>
-        <xsl:with-param name="default.id" select="'title'"/>
-      </xsl:call-template>
-
       <xsl:apply-templates select="title"/>
       <xsl:if test="version">
         <xsl:text> </xsl:text>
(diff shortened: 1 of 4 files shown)


spec/xmlspec-ttml1.xsl
```diff
@@ -612,6 +608,44 @@ width: 100%;
...
+
+<!-- ghrepoloc: participate section -->
+<!-- called in a <dl> content from header -->
+<xsl:template match="ghrepoloc">
+  <xsl:apply-templates/>
+</xsl:template>
+<xsl:template match="ghloc">
+  <dt>Participate:</dt>
+  <dd><a href="https://github.com/{@repo}/">GitHub <xsl:value-of select="@repo"/></a></dd>
+  <dd><a href="https://github.com/{@repo}/issues/">File a bug</a></dd>
```
nigelmegitt: This goes to the issues list not the "create an issue" page - is that intentional?

spec/xmlspec-ttml1.xsl
```diff
@@ -612,6 +608,44 @@ width: 100%;
...
+  <dd><a href="https://github.com/{@repo}/commits/{@branch}">Commit history</a></dd>
+  <dd><a href="https://github.com/{@repo}/pulls/">Pull requests</a></dd>
```
nigelmegitt: This is too much for my tastes: it takes 4 lines that all basically take the reader to the GitHub repo. Is there a reason we need the lines other than the GitHub repo link?

nigelmegitt (commented):
Verified that the CR Snapshot change to entitiescr.dtd works.

spec/xmlspec-ttml1.xsl
himorin: lines are from what respec inserts (e.g. https://w3c.github.io/mediacapture-main/), and I'm not sure what was the original intent...
Actually, bikeshed inserts a link pointing issues/new, so might be better to raise this to respec also?

nigelmegitt: Makes sense to have consistency, yes. Anyway I think we can go straight for issues/new since the text says "File a bug".

nigelmegitt (approved):

